PR #360, opened 2014-12-29
Add invisible username fields to create and login

dandorman:
Functionality is unchanged (the user only sees the password field), but the invisible username field coerces Safari into prompting to save the password.

### DIFF
--- a/app/public/css/styles.css
+++ b/app/public/css/styles.css
@@ -25,6 +25,10 @@ code {
   margin: 0 auto;
 }
 
+.hidden {
+  display: none;
+}
+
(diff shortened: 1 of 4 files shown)


swanson:
@dandorman latest changeset looks good to me - is this ready to go?

Koronen:
I had a suspicion that the input fields might require a `type` attribute to be valid HTML5, but that doesn't seem to be the case. LGTM! :+1:

dandorman:
Yep, I've been running it on my instance and it works as advertised. Merge at your convenience.

PR #361, opened 2015-01-07
Adding password reset

anandubajith:


docs/OpenShift.md
```diff
@@ -116,3 +116,12 @@ After importing feeds, a cron job is needed on OpenShift to fetch feeds.
  ```
 
 5. Done! The cron job should fetch feeds every hour.
+
+Password Reset
+--------------
+In the event that you need to change your password, run the following commands
+```
+rhc ssh app-name
```
Koronen: Maybe this should say `rhc ssh feeds` to match line 10?

anandubajith: yes, but what if people may use another app name

swanson:
Thanks!

PR #363, opened 2015-01-28
Set APP_URL for Heroku

milligramme:
When I created new app on Heroku and set APP_URL, I got a following error.

```
 !    Usage: heroku config:set KEY1=VALUE1 [KEY2=VALUE2 ...]
 !    Must specify KEY and VALUE to set.
```

`heroku apps:info` recently give me two urls, Git URL and Web URL.

ex.
Git URL:       https://git.heroku.com/TESTAPP-cedar-14.git
Web URL:       https://TESTAPP-cedar-14.herokuapp.com/

To set APP_URL I modified regexp to get only Web URL.
Is there more better way to do?

### DIFF
--- a/docs/Heroku.md
+++ b/docs/Heroku.md
@@ -4,7 +4,7 @@ cd stringer
 heroku create
 git push heroku master
 
-heroku config:set APP_URL=`heroku apps:info | grep -o 'http[^"]*/$'`
+heroku config:set APP_URL=`heroku apps:info --shell | grep web_url | cut -d= -f2`
 heroku config:set SECRET_TOKEN=`openssl rand -hex 20`
 
 heroku run rake db:migrate
--- a/docs/Heroku.md
+++ b/docs/Heroku.md
@@ -4,7 +4,7 @@ cd stringer
 heroku create
 git push heroku master
 
-heroku config:set APP_URL=`heroku apps:info | grep -o 'http[^"]*'`
+heroku config:set APP_URL=`heroku apps:info | grep -o 'http[^"]*/$'`
 heroku config:set SECRET_TOKEN=`openssl rand -hex 20`
 
 heroku run rake db:migrate


docs/Heroku.md
```diff
@@ -4,7 +4,7 @@ cd stringer
 heroku create
 git push heroku master
 
-heroku config:set APP_URL=`heroku apps:info | grep -o 'http[^"]*'`
+heroku config:set APP_URL=`heroku apps:info | grep -o 'http[^"]*/$'`
```
Koronen:
```sh
heroku apps:info --shell | grep web_url | cut -d= -f2
```
might be a more reliable way to find the app URL?

docs/Heroku.md
```diff
@@ -33,4 +33,4 @@ heroku restart
 
 ## Password Reset
 
-In the event that you need to change your password, run `heroku run rake change_password` from the app folder.
\ No newline at end of file
+In the event that you need to change your password, run `heroku run rake change_password` from the app folder.
```
Koronen: :+1:

docs/Heroku.md
milligramme: @Koronen thank you. GREAT
I didnt ever know `--shell` option (I needed to update heroku client)
may I send new commit with this reliable way ?

docs/Heroku.md
```diff
@@ -4,7 +4,7 @@ cd stringer
 heroku create
 git push heroku master
 
-heroku config:set APP_URL=`heroku apps:info | grep -o 'http[^"]*'`
+heroku config:set APP_URL=`heroku apps:info --shell | grep web_url | cut -d= -f2`
```
Koronen: I'm not sure when the `--shell` option was introduced. However, seeing as recent versions of the Heroku toolbelt include automatic updates, I don't think backwards comparability will be a problem.

Koronen:
LGTM! :+1:

PR #366, opened 2015-02-17
support for relative base url

posativ:
Tested locally with two configurations, mount to / and mount to /rss. I've never programmed in Ruby/Sinatra, I've tried to match the current style (single-letter helper function for templates).

--

From the commit message:

* enable prefixed redirects so all redirects use the local application
* use uri(..., false, true) to handle \<a href="..."\> locations
* make CSS asset references relative

To run Stringer on a different base url (or url prefix), use

```ruby
require "./app"
map "/rss"do
  run Stringer
end
```

in your 'config.ru'.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -34,6 +34,7 @@ class Stringer < Sinatra::Base
     set :session_secret, ENV["SECRET_TOKEN"] || "secret!"
     enable :logging
     enable :method_override
+    enable :prefixed_redirects
 
     ActiveRecord::Base.include_root_in_json = false
   end
@@ -55,6 +56,10 @@ class Stringer < Sinatra::Base
 
     def t(*args)
       I18n.t(*args)
+    end
+
+    def u(addr)
+      uri(addr, false, true)
     end
   end
 
--- a/app/helpers/authentication_helpers.rb
+++ b/app/helpers/authentication_helpers.rb
@@ -9,6 +9,7 @@ module Sinatra
     end
 
     def needs_authentication?(path)
+      path = uri(nil, false, false)
       return false if ENV['RACK_ENV'] == 'test'
       return false if !UserRepository.setup_complete?
       return false if path == "/login" || path == "/logout"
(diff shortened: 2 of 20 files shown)


swanson:
Wow, thanks so much - this looks promising.

It might be a week or so until I have time to fully review and test this out. Are you currently running this on Heroku? Or another platform?

Koronen:
Ref: #288

posativ:
I am hosting it on a virtual server, so basically the bundler installation/execution.

Didn't see #288, but my PR only proposes relative URLs, so basically `url(addr, false, true)` (defaults to `url(addr, true, true)` IIRC). It uses `PATH_INFO` then which makes it possible to run the application on a different mount point (as per Rack specification).

Sinatra/AssetPack runs just fine, didn't encountered any trouble.

posativ:
I've encountered a missing form location and the session gets cleared on POST requests. I've fixed it locally, but I'll think about a way to handle session properly on non-root locations.

swanson:
I'm also curious if this will still work with the Fever-compatible API (normally mounted to `/fever`) - should this use the new mount point as well?

posativ:
> I'm also curious if this will still work with the Fever-compatible API (normally mounted to /fever) - should this use the new mount point as well?

The Fever-API is just another rack application and can be mounted to /fever if you run two processes. I didn't test a Fever-compatible mobile application – if the application is closed-source and doesn't support a relative endpoint, it doesn't work of course.

Everything else can be patched.

---

The only issue now is the JavaScript application: `/stories` and `/feeds` can't be relative. There is a fix for this (traversing the DOM for `script` elements and check if the filename looks like the application script and compute the used url from that).

swanson:
Sure - what I meant about `/fever` was can it still be mounted to e.g `/rss/fever` if you mount Stringer at `/rss`.

posativ:
Yes.

swanson:
Hey @posativ -- sorry I've dropped the ball on this. Could you refresh my memory on the outstanding issues? Looks like there might be some JavaScript problems to sort out still?

Koronen:
Closing this due to lack of activity. Feel free to re-open if there are any updates on this.

PR #376, opened 2015-04-30
Fill feed URL field from params

Koronen:
Read the feed URL from params on request to `/feeds/new` and let it fill the feed URL field. This enables integration with the [RSS Subscription Extension][rss-subscription-ext] extension for Google Chrome and should allow integration with similar extensions, for Google Chrome or other browsers.

[rss-subscription-ext]: https://chrome.google.com/webstore/detail/rss-subscription-extensio/nlbjncdgjeocebhnmkbbbdekmmmcbfjd

### DIFF
--- a/app/controllers/feeds_controller.rb
+++ b/app/controllers/feeds_controller.rb
@@ -31,6 +31,7 @@ class Stringer < Sinatra::Base
   end
 
   get "/feeds/new" do
+    @feed_url = params[:feed_url]
     erb :'feeds/add'
   end
 


swanson:
What do you think about a different endpoint? Just so it's clear how this could be used `/feeds/oneclick` or something?

swanson:
Looks like a neat extension - does it open the page in Stringer and we have to click 'Add'? Or is the intent that we can subscribe directly from the `GET`?

Koronen:
When using the extension to subscribe to a feed, it opens the feed reader in a new tab. Feeds and corresponding URLs are configurable, but only `GET`-requests appear to be supported.

My first thought was to add a new endpoint to add the feed directly, without user interaction. Then I changed my mind and implemented it this way because it's the minimal functional patch and let's us avoid using `GET` requests to modify application state.

swanson:
:+1: works for me

PR #381, opened 2015-08-12
use TLS for external resources

InPermutation:
3 changes, in decreasing order of importance:

* Heroku automatically sends you to the HTTPS version of the app when you install with Heroku Button.
Then, Firefox (since [version 23](https://developer.mozilla.org/en-US/docs/Security/MixedContent/How_to_fix_website_with_mixed_content)) blocks the download of Reenie Beanie, and apparently replaces it with Comic Sans MS?!
* The changes in README.md are mostly irrelevant because of https://help.github.com/articles/why-do-my-images-have-strange-urls/, but it's good practice, in case it ends up getting rendered somewhere that isn't GitHub.
* I'm not sure if the Placekitten one is even viewable from a browser -- I didn't bother checking. The HTTPS version of the resource does load, however, so it probably won't break anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # Stringer
 
-[![Build Status](http://img.shields.io/travis/swanson/stringer.svg?style=flat)](https://travis-ci.org/swanson/stringer)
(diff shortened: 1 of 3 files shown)


Koronen:
Thanks @InPermutation!

PR #387, opened 2015-10-05
Redirect to the previous page after login

aliou:
When going to an URL that needs authentication (like `/archive/` or `/feed/:id`) we currently get redirected to the login page and then to the feed index. This redirects to the page we initially tried to access after login.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -10,11 +10,8 @@ class Stringer < Sinatra::Base
     if user
       session[:user_id] = user.id
 
-      if session[:redirect_to].present?
-        redirect to(session.delete(:redirect_to))
-      else
-        redirect to("/")
-      end
+      redirect_uri = session.delete(:redirect_to) || '/'
+      redirect to(redirect_uri)
     else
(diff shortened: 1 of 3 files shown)


Koronen:
LGTM! :+1:

app/controllers/sessions_controller.rb
```diff
@@ -10,7 +10,11 @@ class Stringer < Sinatra::Base
     if user
       session[:user_id] = user.id
 
-      redirect to("/")
+      if session[:redirect_to].present?
```
pascalw: Small opportunity to get rid of the conditional here, by specifying a default `redirect_to` like so:

```ruby
redirect_uri = session.delete(:redirect_to) || '/'
redirect to(redirect_uri)
```

Alternatively you could use the block syntax to specify a default, like so:

```ruby
session.delete(:redirect_to) { '/' }
```

I personally prefer the first, without block.

aliou: Good call, thank you!

swanson:
:+1: thanks @aliou :)

PR #391, opened 2015-12-07
Rubocop and code formatting

jeroenj:
I've generated a `rubocop_todo.yml` and a `rubocop.yml`. This currently fixes only spacing related offences.

If I find the time and there is interest I might look into fixing some other offences.


Koronen:
LGTM! :+1: I'm all for adding a style guide to this project. @swanson?

swanson:
:+1: to adding it. But I'd rather fix the outstanding issues before merging it in with a ton of exceptions :)

I think this repo is low volume enough that it should be doable.

@jeroenj I'm curious if there is a baseline style guide that these rules are based on? I don't have strong opinions but would prefer one with some existing community buy-in.

Koronen:
RubyCop's default configuration is based on the community-driven "The Ruby Style Guide", https://github.com/bbatsov/ruby-style-guide/.

Koronen:
> But I'd rather fix the outstanding issues before merging it in with a ton of exceptions :)

While I would also prefer to have all the warnings fixed now, I can imagine that taking quite some time and effort. I'm open to the option of merging this PR as-is and fixing the warnings in a series of smaller PRs, which I think could be easier for the author(s) to write and for the reviewer(s) to review.

jeroenj:
Yes, I'd merge it as it is right now too. Since all the exceptions are in a todo file it's quite clear that they're pending. I've also added the project to [RefactorCop](http://www.refactorcop.com/swanson/stringer) so it might attract other people willing to contribute.

I generally try to use the default config as much as possible. I noticed that some of the code has different styling so having separate pull requests might for some changes might be easier to open the discussion.

I'm willing to go through all of them (I already did it several times for other projects) but it might take some time to get it done. :-)

swanson:
I don't have a strong opinion, so merge away :) 

Can we add style check to the build job?

Koronen:
> Can we add style check to the build job?

Yes, please. :smiley:  Do we want to do it in a separate PR or in this one. I'm fine with either.

jeroenj:
Yes, I'll look into it. But not right now. 😃 (it's getting late in CET 😄)

jeroenj:
@swanson I've added it to the build config. The build will fail if rubocop detects offences.

I noticed that ruby 2.1 and 2.2 builds are allowed to fail? Shouldn't we remove that requirement? If one of those would fail it would not be noticed, right?

swanson:
@jeroenj yeah, that's fine if you want to change that

jeroenj:
@swanson I've changed the build setup to not allow failures on 2.1 and 2.2.

If all looks good to you guys this might be merged. I don't think that I'll have the time to go through the other offences in the next few days. By merging there might be a (slight? 🙂) chance that somebody else starts working on it.

Koronen:
LGTM! :+1:

PR #392, opened 2015-12-08
Fix RuboCop lint offences

Koronen:
~~There's currently one rule left, `Lint/RescueException`, that needs more thought.~~

Ready for review.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -13,12 +13,6 @@ Lint/RescueException:
     - 'app/utils/feed_discovery.rb'
 
 # Offense count: 5
-Lint/UselessAssignment:
-  Exclude:
-    - 'app/tasks/change_password.rb'
-    - 'spec/repositories/feed_repository_spec.rb'
-
-# Offense count: 5
 Metrics/AbcSize:
   Max: 37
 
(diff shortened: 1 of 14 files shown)


swanson:
Those exceptions could be scope to whatever Feedjira can create; but I'm not aware of a great way to get that list.

swanson:
:+1: thanks for tackling that :)

Koronen:
Thanks for the hint! I'll have a look next time I have a chance to work on this.

swanson:
Let's merge for now with the inline exceptions -- I can live with those :)

Koronen:
* Removed the unused `finder` parameter in c190d3c0bff7a297363a4dea4f5ca34adc4b8172
* Fixed Lint/RescueException in 0ae414410d7072bcbae6524c81efff7d31279fe8.

  This was done by rescuing `StandardError` instead of `Exception`, which should still cover most if not all errors raised by the Curb and Feedjira libraries. We no longer try to rescue fatal errors, like `NoMemoryError`, which is what RuboCop was complaining about.

  An even better fix for this could be to rescue an even narrower set of errors, but I'll leave that until the next time.

swanson:
Thanks!

PR #393, opened 2015-12-18
Adds group_id to Feeds#edit

jeroenj:
I've been using groups for quite a while. While this is a hidden feature it would be handy that feeds could be linked using the interface. The dropdown will only be shown when there are any groups so that people not using the feature don't get confused about it.

I might also look into more group management later on, but I rarely use the web interface so it's not really a priority for me right now. (and I've got a lot of other stuff on my plate)

The only think that needs to be fixed is the design of the dropdown. I guess stringer is a custom theme? Would any of you mind to help me fix it? I'll try to fix it myself but it might take a few days/weeks.

screen shot 2015-12-18 at 22 03 20

### DIFF
--- a/app/controllers/feeds_controller.rb
+++ b/app/controllers/feeds_controller.rb
@@ -18,7 +18,7 @@ class Stringer < Sinatra::Base
   put "/feeds/:id" do
     feed = FeedRepository.fetch(params[:id])
 
-    FeedRepository.update_feed(feed, params[:feed_name], params[:feed_url])
+    FeedRepository.update_feed(feed, params[:feed_name], params[:feed_url], params[:group_id])
 
     flash[:success] = t('feeds.edit.flash.updated_successfully')
     redirect to('/feeds')
--- a/app/public/css/styles.css
+++ b/app/public/css/styles.css
@@ -489,9 +489,17 @@ li.feed .remove-feed a:hover {
   transition: 0.25s;
 }
 
-.setup #password, .setup #password-confirmation, .setup #feed-url, .setup #feed-name {
+.setup #password, .setup #password-confirmation, .setup #feed-url, .setup #feed-name, .setup input.select-dummy {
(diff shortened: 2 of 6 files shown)


Koronen:
I myself don't use the groups feature, but this looks like a reasonable addition IMO.

swanson:
Not sure on this...don't use groups myself either. I think it will be confusing to people using the app to see UI elements for editing groups, but no changes to the rest of the app.

jeroenj:
Yes, that's why the group field is only shown if you have any groups defined. People that are not using the groups feature won't see this field.

pascalw:
I do use groups so I think this would definitely be useful. Hiding the drop down for people that don't have any groups defined seems reasonable to me. Perhaps in the future we could add a simple crud interface for groups to make it even more useful.

jeroenj:
Yes, but I think a CRUD for groups would only make sense if they're actually used in the web app itself. As long as that is not the case I'd hide it with a config so it won't confuse people that only use the web app.

jeroenj:
I took a stab at the design and ended up with this:

screen shot 2015-12-23 at 22 15 42
screen shot 2015-12-23 at 22 15 51

I've also cleaned up the commit a bit.

pascalw:
@jeroenj code looks fine :+1:.

However I just took your branch for a spin and noticed that the current group of a feed is not correctly set on the edit feed page. Looking at the code you're missing something to mark an `option` `selected`.

jeroenj:
Yes, indeed. I found out too. I did not have the time yet to fix it.

jeroenj:
@pascalw I'v fixed marking the current `option` as `selected`. Let me know if you find any other issues. :)

pascalw:
@jeroenj works fine now :+1: 

@swanson @Koronen Are you ok merging this?

Koronen:
LGTM! :+1:

PR #395, opened 2015-12-19
Fix RuboCop metrics offences

Koronen:
* Fix Metrics/AbcSize

  ...through some light refactorings.

* Fix Metrics/CyclomaticComplexity

  ...through a small refactoring of `Sinatra::AuthenticationHelpers#needs_authentication?`.

* Set Metrics/LineLength maximum to 120 characters

  Matt and I agreed that 120 characters is a sensible maximum line length for modern editors, see
<https://github.com/swanson/stringer/pull/395#issuecomment-166069080>.

  Also set the `MaxLineLength` option for two other cops, since they're related.

  Joins lines in the Fever API files to fix warnings from `Style/IfUnlessModifier`.

* Lower Metrics/MethodLength maximum to 15 lines

  Despite my best efforts, I find it hard to keep all methods below or at the default maximum length of 10 lines. 15 lines is a somewhat arbitrary compromise that's subject to future tweaking.

  See also: <https://robots.thoughtbot.com/sandi-metz-rules-for-developers>

* Fix Metrics/PerceivedComplexity

* Fix Metrics/ClassLength

  Fixes Metrics/ClassLength by extracting `StoryRepository#expand_absolute_urls` and `StoryRepository#normalize_url` to a separate helper module.

### DIFF
--- a/app/commands/feeds/find_new_stories.rb
+++ b/app/commands/feeds/find_new_stories.rb
@@ -15,10 +15,8 @@ class FindNewStories
 
     @raw_feed.entries.each do |story|
       break if @latest_entry_id && story.id == @latest_entry_id
-
-      unless story_age_exceeds_threshold?(story) || StoryRepository.exists?(story.id, @feed_id)
-        stories << story
-      end
+      next if story_age_exceeds_threshold?(story) || StoryRepository.exists?(story.id, @feed_id)
+      stories << story
     end
 
     stories
--- a/app/helpers/authentication_helpers.rb
+++ b/app/helpers/authentication_helpers.rb
@@ -11,9 +11,8 @@ module Sinatra
     def needs_authentication?(path)
       return false if ENV['RACK_ENV'] == 'test'
       return false if !UserRepository.setup_complete?
-      return false if path == "/login" || path == "/logout"
-      return false if path =~ /css/ || path =~ /js/ || path =~ /img/
-      return false if path == "/heroku"
+      return false if %w(/login /logout /heroku).include?(path)
+      return false if path =~ /css|js|img/
       true
     end
 
--- a/app/repositories/story_repository.rb
+++ b/app/repositories/story_repository.rb
@@ -1,7 +1,10 @@
+require_relative "../helpers/url_helpers"
 require_relative "../models/story"
 require_relative "../utils/sample_story"
 
 class StoryRepository
+  extend UrlHelpers
+
   def self.add(entry, feed)
     entry.url = normalize_url(entry.url, feed.url)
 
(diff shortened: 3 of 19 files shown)


Koronen:
For some reason, extracting `StoryRepository#expand_absolute_urls` and `StoryRepository#normalize_url` breaks the build. I don't (yet) know why, so I'd appreciate some help.

app/helpers/authentication_helpers.rb
```diff
@@ -9,12 +9,13 @@ def is_authenticated?
...
-      return false if path =~ /css/ || path =~ /js/ || path =~ /img/
-      return false if path == "/heroku"
-      true
+      [
+        ENV['RACK_ENV'] == 'test',
+        !UserRepository.setup_complete?,
+        path == '/login' || path == '/logout',
+        path =~ /css/ || path =~ /js/ || path =~ /img/,
+        path == '/heroku'
+      ].none?
```
swanson: I find this to be less readable personally. Any other ideas?

app/repositories/story_repository.rb
```diff
@@ -1,7 +1,10 @@
+require_relative "../helpers/url_helpers"
 require_relative "../models/story"
 require_relative "../utils/sample_story"
 
 class StoryRepository
+  extend UrlHelpers
```
swanson: Maybe failing test could be due to `extend` vs `include`? Just a stab in the dark :)

swanson:
>Fix Metrics/LineLength
>Breaks long lines to conform to the RuboCop default maximum width of 80 characters.

I think I'd prefer a longer line length setting if I'm being honest. 80 seems small for modern editors/resolutions and I think a lot of the changes that move code to multiple lines hurt readability.

What do you think?

swanson:
> Lower Metrics/MethodLength maximum to 15

I'm fine with this instead of trying to fit everything under 10

app/helpers/authentication_helpers.rb
Koronen: How about the following?
```rb
def needs_authentication?(path)
  return false if ENV['RACK_ENV'] == 'test'
  return false if !UserRepository.setup_complete?
  return false if %(/login /logout /heroku).include?(path)
  return false if path =~ /css|js|img/
  true
end
```

app/repositories/story_repository.rb
Koronen: Switching to `include UrlHelpers` adds `expand_absolute_urls` and `normalize_url` as instance methods (instead of class methods) and causes `NoMethodError: undefined method 'expand_absolute_urls' for StoryRepository:Class`. Unfortunately, that doesn't appear to be it. :worried:

Koronen:
> I think I'd prefer a longer line length setting if I'm being honest. 80 seems small for modern editors/resolutions and I think a lot of the changes that move code to multiple lines hurt readability.
>
> What do you think?

I'm perfectly fine with increasing the maximum line length. How about setting it to 120 characters?

> > Lower Metrics/MethodLength maximum to 15
>
> I'm fine with this instead of trying to fit everything under 10

Great! :smiley:

app/helpers/authentication_helpers.rb
swanson: :+1:

swanson:
120 works for me :)

Koronen:
This PR is now ready for another round of review.

* `Sinatra::AuthenticationHelpers#needs_authentication?` has been updated according to feedback.
* `Metrics/LineLength` maximum has been set to 120 characters, as agreed upon.
* The breaking build has been fixed. It was caused by an accidentally inverted conditional in `UrlHelpers#expand_absolute_urls`.

app/commands/feeds/find_new_stories.rb
```diff
@@ -16,9 +16,7 @@ def new_stories
     @raw_feed.entries.each do |story|
       break if @latest_entry_id && story.id == @latest_entry_id
 
-      unless story_age_exceeds_threshold?(story) || StoryRepository.exists?(story.id, @feed_id)
-        stories << story
-      end
+      stories << story unless story_age_exceeds_threshold?(story) || StoryRepository.exists?(story.id, @feed_id)
```
Koronen: I'm thinking extracting the conditional to a guard clause might improve readabilty. Thoughts?
```rb
@raw_feed.entries.each do |story|
  break if @latest_entry_id && story.id == @latest_entry_id
  next if story_age_exceeds_threshold?(story) || StoryRepository.exists?(story.id, @feed_id)
  stories << story
end
```

PR #396, opened 2015-12-19
Fix RuboCop performance offences

Koronen:
* Fix Performance/StringReplacement

  > ### delete([other_str]+) -> new_str
  >
  > Returns a copy of *str* with all characters in the intersection of its
  > arguments deleted. Uses the same rules for building the set of characters as
  > `String#count`.

  See <http://ruby-doc.org/core-2.0.0/String.html#method-i-delete>.


swanson:
Better performance and I think it actually is more clear :) :+1:

PR #397, opened 2015-12-19
Fix RuboCop style offences

Koronen:
Fixes all remaining style offences reported by RuboCop.

TODO:
- [x] Wait for #395 to be resolved
- [x] Wait for #396 to be resolved
- [x] Rebase!
- [x] Fix Style/NumericLiterals
- [x] Fix Style/StringLiterals


Koronen:
I'd like some feedback on how to fix the two remaining style offences.

* Style/NumericLiterals

  Most of the offences reported by this cop are UNIX timestamps. I'm not sure if adding underscores improves readability in these cases.

* Style/StringLiterals

  Enforcing one of the two supported styles would result in *a lot* of changes pretty much everywhere. Do we want to enforce single quotes (RuboCop default), enforce double quotes or just turn off this cop? I'd prefer enforcing one of the two options for consistency, but I don't have a strong opinion on which one. If I had to choose I'd go for single quotes because it's the RuboCop default.

swanson:
> Style/NumericLiterals

Yeah I don't think the underscore here is the right answer -- I don't know if I would want that rule even for non-timestamp large numbers. +1 to disabling that.

>Style/StringLiterals

I personally type double-quotes as my 'default'. I agree that consistency is for the best. One small thing that has me favoring double-quotes is that can't use `#{foo.bar}` style interpolation with single-quotes. But, hey, you're the one doing all the heavy lifting so I'm okay with your preference :)

Koronen:
> I personally type double-quotes as my 'default'. I agree that consistency is for the best. One small thing that has me favoring double-quotes is that can't use `#{foo.bar}` style interpolation with single-quotes.

Let's enforce double quotes then?

> But, hey, you're the one doing all the heavy lifting so I'm okay with your preference :)

My preference is not a strong one, and it will be RuboCop doing all the heavy lifting in this case. This cop supports [auto-correction](https://github.com/bbatsov/rubocop/wiki/Automatic-Corrections). :smiley:

swanson:
:+1:

PR #406, opened 2016-01-29
Fix outdated documentation

guiniol:
`rubocop` now has a `require` so we need to change it to the old syntax for openshift.


swanson:
Thanks!

guiniol:
No worries. Just helping the next person with that same problem ^^

PR #408, opened 2016-02-04
Update to new/supported ruby version.

geemus:

Issue #364, opened 2015-01-30
ruby 2.2


swanson:
- [x] Check if upgrade works as expected
- [x] Add 2.2 to travis build matrix
- [x] Bump .ruby-version?
- [ ] Update deployment instructions (VPS, openshift, heroku)

duenni:
Could this be an issue for Openshift users? You can only use Ruby 2.0 there...

swanson:
Potentially, will add to checklist

jeroenj:
It may not be 2.2 but I've been running it on 2.1.x for quite some time now. I'm planning to upgrade to 2.2 soon. I'll let you know (and/or provide PR's) if I find any issues. :)

Koronen:
Ruby 2.2 was added to the Travis CI build matrix in #365.

Koronen:
Sadly, the [Ruby "Overview" page on OpenShift](https://developers.openshift.com/en/ruby-overview.html) (at the time of this writing) still states that they only support Ruby 2.0 from the Ruby 2.x branch. :disappointed:

bf4:
Any reason you lock the version to 2.0.0 in the Gemfile? That makes sense in your company, but in an OSS unless it is absolutely required, it could be just a recommendation or step in setup.

swanson:
> Any reason you lock the version to 2.0.0 in the Gemfile? That makes sense in your company, but in an OSS unless it is absolutely required, it could be just a recommendation or step in setup.

@bf4 sorry for the late response here. Heroku advises adding a Ruby version to your Gemfile so that the deployed environment is explicitly defined. Since Heroku is the 'official' recommended deployment story, I've opted to add this as a convenience for those users. We want a smooth installation process, which means that we are a little more restrictive in the Ruby version that we recommend.

The trade-off is that folks deploying off the main path will have a little more work to do, but that is a decision that I decided I could live with :)

bf4:
I have an incomplete PR which basically makes it configurable, but default to 2.0.0.  Would that be of interest?

swanson:
> I have an incomplete PR which basically makes it configurable, but default to 2.0.0. Would that be of interest?

Happy to take a look at it :) As long as the Heroku considerations I mentioned above are satisfied, I'm open to it.

dwaite:
Would this help with open shift users? https://github.com/openshift/sti-ruby

moritzheiber:
[...]

PR #410, opened 2016-02-05
RFC: Lots of ways to configure the Ruby version

bf4:
1. In a `Gemfile.local`
1. From the environment `$RUBY_VERSION` (not sure how this behaves on JRuby)
1. From the `.ruby-version` file
1. Defaults to `2.0.0`

Ref:

- https://github.com/swanson/stringer/issues/364
- https://github.com/swanson/stringer/pull/408

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ rvm:
   - 2.0.0
   - 2.1
   - 2.2
+  - 2.3.0
 script:
   - bundle exec rspec
   - mocha-phantomjs http://localhost:4567/test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 before_install:
   - gem update bundler
-  - sed -i '1d' Gemfile
+  - 'echo "RUBY_VERSION: $RUBY_VERSION"'
 before_script:
   - npm install -g mocha-phantomjs@2.0.2
   - bundle exec rake test_js &> /dev/null &
--- a/Gemfile
+++ b/Gemfile
@@ -1,19 +1,5 @@
-# Add a Gemfile.local to locally bundle gems outside of version control
-# and/or specify alternate Ruby versions to bundle
-local_gemfile = File.join(File.expand_path("..", __FILE__), "Gemfile.local")
-if File.readable?(local_gemfile)
-  eval_gemfile local_gemfile
-else
-  ruby_version = ENV.fetch("RUBY_VERSION") do
-    ruby_version_file = File.join(File.expand_path("..", __FILE__), ".ruby-version")
-    if File.readable?(ruby_version_file)
-      File.read(ruby_version_file)
-    else
-      "2.0.0"
-    end
-  end.split("-")[-1] # e.g. {jruby,ruby,rbx}-2.0.0 -> 2.0.0
-  ruby ruby_version
-end
+ruby_version_file = File.join(File.expand_path("..", __FILE__), ".ruby-version")
+ruby File.read(ruby_version_file).chomp if File.readable?(ruby_version_file)
 source "https://rubygems.org"
 
 group :production do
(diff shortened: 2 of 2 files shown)


bf4:
CI checks appear to be a connection problem, will need to be re-run when it resolves

.travis.yml
```diff
@@ -1,6 +1,6 @@
 before_install:
   - gem update bundler
-  - sed -i '1d' Gemfile
+  - echo "RUBY_VERSION: $RUBY_VERSION"
```
Koronen: I restarted the builds, but they still fail. This time it appears to be because of a string quoting issue on this line.
```
$ {:"echo \"RUBY_VERSION"=>"$RUBY_VERSION\""}
/home/travis/build.sh: line 45: {:echo "RUBY_VERSION=: command not found
```
To fix it

Gemfile
```diff
@@ -1,4 +1,19 @@
-ruby "2.0.0"
+# Add a Gemfile.local to locally bundle gems outside of version control
+# and/or specify alternate Ruby versions to bundle
+local_gemfile = File.join(File.expand_path('..', __FILE__), 'Gemfile.local')
+if File.readable?(local_gemfile)
+  eval_gemfile local_gemfile if File.readable?(local_gemfile)
```
Koronen: Any reason why we check `File.readable?(local_gemfile)` twice?

Gemfile
```diff
@@ -1,4 +1,19 @@
-ruby "2.0.0"
+# Add a Gemfile.local to locally bundle gems outside of version control
+# and/or specify alternate Ruby versions to bundle
+local_gemfile = File.join(File.expand_path('..', __FILE__), 'Gemfile.local')
```
Koronen: Minor thing, but our [recently introduced RuboCop configuration](https://github.com/swanson/stringer/blob/7fbfce8d84c2d9be1903c93a9c432b4c7e0825a9/.rubocop.yml#L24-L25) says use double quotes.

Gemfile
```diff
@@ -1,4 +1,19 @@
...
+  eval_gemfile local_gemfile if File.readable?(local_gemfile)
+else
+  ruby_version = ENV.fetch("RUBY_VERSION") {
+    ruby_version_file = File.join(File.expand_path('..', __FILE__), '.ruby-version')
+    if File.readable?(ruby_version_file)
+      File.read(ruby_version_file)
+    else
+      "2.0.0"
+    end
+  }.split('-')[-1] # e.g. {jruby,ruby,rbx}-2.0.0 -> 2.0.0
```
Koronen: `[-1]` -> `.last`?

Koronen:
I like the options that this PR introduces for selecting a Ruby version and add local gems. :+1: Some thoughts:

* What would you say about breaking the outer conditional statement, allowing you to select a Ruby version from `ENV["RUBY_VERSION"]` or `.ruby-version`, _and_ load local gems from `Gemfile.local`? That would be more flexible?
* Regarding `File.readable?` vs. `File.exist?`: I don't mind using the more restrictive `File.readable?`.
* I like to see us move forward with this PR once the build is green.

swanson:
My preference would be to have one place of defining the version. `.ruby-version` would be my suggestion since it will support both `rvm` and `rbenv`, as well as being part of the codebase directly. If we want to add executable code to the Gemfile, it could output the version from `.ruby-version` at the top to appease Heroku.

If someone wants to deploy with a different ruby version, I think modifying `.ruby-version` is a straightforward and reasonable thing to do. Having multiple ways to pick the version could cause confusion about priority/order (does env var trump a Gemfile.local? etc) and will complicate any instructions we write.

bf4:
Ok, updated to the `.ruby-version` only version.  The  one with the global, at least on my machine, that global is just there

bf4:
I rebased off of master, and while at it, cleaned up the commits, so some of the comments about typos are out of date :)

Koronen:
LGTM! :+1:

swanson:
So my thinking is that now we can remove `.ruby-version` from `.gitignore` and and commit a `.ruby-version` with 2.0.0 (or 2.3 when we want to flip that switch), does that make sense?

bf4:
Just depends on if people should change source control

B mobile phone

Issue #411, opened 2016-02-09
Improve robustness of rake work_jobs


zlogic:
First of all, many thanks on creating this awesome RSS reader! Just what I was looking for and was about to create myself, but Stringer is so much better than anything I wanted :)

I installed it into Heroku and noticed a problem with reliability of the Heroku-specific `rake lazy_fetch` approach:

While the `bundle exec rake work_jobs` is launched from config/unicorn.rb, it's never checked to be running and its delayed_job_pid is never used. It seems that Heroku kills long-running database connections either on purpose, or just randomly, which also causes an unhandled exception to kill the `bundle exec rake work_jobs` process ([example log](https://github.com/swanson/stringer/files/123722/ConnectionLoss.txt)). While `rake lazy_fetch` continues to submit jobs, the worker process is dead and never restarted.

It seems the best way would be to run `rake work_jobs` [from a Procfile](http://stackoverflow.com/a/14855937/2401011). I've tried two other workarounds and can confirm they work:

1. Restart the rake process in bash:
```
diff --git config/unicorn.rb config/unicorn.rb
--- config/unicorn.rb
+++ config/unicorn.rb
@@ -9,7 +9,7 @@ before_fork do |_server, _worker|
   # as there's no need for the master process to hold a connection
   ActiveRecord::Base.connection.disconnect! if defined?(ActiveRecord::Base)
 
-  @delayed_job_pid ||= spawn("bundle exec rake work_jobs")
+  @delayed_job_pid ||= spawn("while true; do bundle exec rake work_jobs; sleep 60; done")
 
   sleep 1
 end
```
2. Catch exceptions in the work_jobs job and try to reconnect:
```
diff --git Rakefile Rakefile
--- Rakefile
+++ Rakefile
@@ -46,11 +46,25 @@ desc "Work the delayed_job queue."
 task :work_jobs do
   Delayed::Job.delete_all
 
-  3.times do
-    Delayed::Worker.new(
-      min_priority: ENV["MIN_PRIORITY"],
-      max_priority: ENV["MAX_PRIORITY"]
-    ).start
+  logger = Logger.new(STDOUT)
+  logger.level = Logger::DEBUG
+
+  loop do
+    begin
+      logger.info "(Re-)starting worker"
+      if !ActiveRecord::Base.connection.active?
+        logger.info "Restarting DB connection"
+        ActiveRecord::Base.connection.reconnect!
+      end
+      Delayed::Worker.new(
[...]

PR #412, opened 2016-02-10
Proposal for fixing issue #411

zlogic:
Simple fix for #411, works well on the free tier of Heroku.


zlogic:
This didn't work well

zlogic:
Heroku billing had a glitch today, sending hundred or even thousand-dollar bills to many customer.
https://status.heroku.com/incidents/854

Looks like their scheduler service had some problems, I got a notice that the scheduler accumulated $701.78 in charges last week!

zlogic:
Will recreate this PR if it didn't cause the crazy billing :)

PR #417, opened 2016-02-23
Update Heroku.md (heroku addons:create)

milligramme:
> WARNING: `heroku addons:add` has been deprecated. Please use `heroku addons:create` instead.


swanson:
Thanks!

PR #419, opened 2016-02-27
Alternative method of running rake work_jobs

zlogic:
A fix for the forked `bundle exec rake work_jobs` process crashing on Heroku (issue #411).
This includes a basic worker (just run `bundle exec rake work_jobs` and restart on failure) and a "full worker" which does most of the daily tasks without a scheduler. It's possible to choose which ones to use in the dashboard and having up to 1 worker is free of charge.

worker

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,10 @@ Lato Font Copyright &copy; 2010-2011 by tyPoland Lukasz Dziedzic (team@latofonts
 
 ## Contact
 
+If you have a question, feature idea, or are running into problems, our preferred method of contact is to open an issue on GitHub. This allows multiple people to weigh in and we can keep everything in one place. Thanks!
+
+## Maintainers
+
 Matt Swanson, [mdswanson.com](http://mdswanson.com), [@_swanson](http://twitter.com/_swanson)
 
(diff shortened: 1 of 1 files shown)


swanson:
Thanks - I need to look into this more in-depth. I missed this blog post: https://blog.heroku.com/archives/2015/5/7/heroku-free-dynos that introduces the free worker dyno, which sounds awesome.

Koronen:
I like the gist of this fix. Thank you! :+1:

Here are some of my thoughs:

* Do we need two process definitions?

  The docs currently instruct users to configure cron jobs (Heroku Scheduler or similar) for scheduling fetching of new stories, and for cleaning out old stories. Maybe it's sufficient to keep just the `delayed_worker` process definition?

* Do we need the loop?

  I think at least Heroku restarts crashed processes automatically.

  > Dynos are also restarted if the processes running in the dyno exit.

  https://devcenter.heroku.com/articles/dynos#restarting

* How can we provide a smooth upgrade process for everyone using this app?

  There's probably some documentation that could be updated to reflect this change in behavior.

gabrielpoca:
Closing this PR for lack of activity.

PR #426, opened 2016-03-10
Improving how data is passed to Delayed Job

fearenales:
This PR intends to make the Delayed Job processing more robust.

In my tests, while running Stringer in a Docker container, there was an issue on passing the `Feed` data to the Delayed Job.

The whole object was serialized by the application (`FetchFeeds` class) and deserialized by the Delayed Job.
That's not a good practice and can cause several issues, as you can see in https://github.com/collectiveidea/delayed_job_active_record/issues/101

To solve this, instead of passing the whole `Feed` object to the Delayed Job, only its `id` attribute is sent. Then, the `Feed` is get from the database and its content is fetched normally.

Plus, the `@pool` object (a pool of threads) was also serialized and sent, causing issues.
If no `pool` is passed to the job, it should be instantiated from inside the job itself, avoiding the serialize/unserialize issues.
Keeping the `pool` argument in the `FetchFeeds#initialize` method ensures the tests can still mock it and run the defined `expect` clauses.

Finally, in order to make it easy to configure the worker component (set how many times it should be respawned; if it should be spawned by this process or it will be invoked elsewhere), couple optional env vars were added.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -46,7 +46,8 @@ desc "Work the delayed_job queue."
 task :work_jobs do
   Delayed::Job.delete_all
 
-  3.times do
+  worker_retry = Integer(ENV["WORKER_RETRY"] || 3)
+  worker_retry.times do
     Delayed::Worker.new(
       min_priority: ENV["MIN_PRIORITY"],
       max_priority: ENV["MAX_PRIORITY"]
--- a/app.json
+++ b/app.json
@@ -23,6 +23,14 @@
     "ENFORCE_SSL": {
       "description": "Force all clients to connect over SSL",
       "value": "true"
+    },
+    "WORKER_EMBEDDED": {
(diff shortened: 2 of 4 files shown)


swanson:
Thanks, I agree with everything you've said. Would love to revisit the background job code, but this looks like a good start. I need to make some time to test this myself, so bear with me :)

Koronen:
Maybe document the new environment variables somewhere? LGTM otherwise! :+1: 

If you feel comfortable working with Git, maybe you could squash the Rubocop fixes into the corresponding commit(s) before, and amend the first commit adding most of the PR description as a git commit message?

fearenales:
@Koronen Thanks for your feedback, I've just improved the commits.

Plus, I've added the env vars description in the `app.json` file (where other env vars were used and documented as well -- and I don't know if you guys would like to have them in the README.md).

PR #428, opened 2016-03-22
Upgrade all the gems

Koronen:
Update version requirements to lock at the minor level by default, except for some pre-1.0 gems used in production. This should allow us to upgrade gems more freely, assuming all gems and dependencies use semantic versioning.

* Remove `arel` from `Gemfile`

  It's already a dependency of `activerecord` and we don't depend on it directly   in any way.

* Remove the version requirement from `i18n`

  We do directly depend on this gem, but for the time being, we can let `activerecord` (and its dependency `activesupport`) dictate the version.

* Disable the `Style/FrozenStringLiteralComment` Cop for now

  It was very noisy.

* Install bcrypt instead of bcrypt-ruby

        #######################################################
        Post-install message from bcrypt-ruby:
        #######################################################
        The bcrypt-ruby gem has changed its name to just bcrypt.  Instead of
        installing `bcrypt-ruby`, you should install `bcrypt`.  Please update your
        dependencies accordingly.

* Add a Gemnasium badge


Koronen:
I've had this branch successfully deployed to my instance for close to a month now without any noticeable issues.

PR #431, opened 2016-03-30
Fixing redirections between pages

fearenales:
This PR intends to improve the redirections between Stringer pages. This should protect inner pages and expose only proper public pages.

If a user hasn't signed up yet, they shouldn't access inner pages (such as `/news`). In this case, those pages and the `/login` page should redirect to `/setup/password`.

If a user has signed up, but still hasn't completed the tutorial step (and they are logged in), inner pages, `/login`, `/setup/password` pages are all redirected to `/setup/tutorial`. The exception is `/feeds/import` page, which can be accessed even if the user hasn't complete the tutorial step.

In the same sense as above, if a user has signed up, but still hasn't completed the tutorial step (and they aren't logged in), the inner pages, `/setup/password` and `/feeds/import` pages are all redirected to `/login`.

If a user has signed up and has completed the tutorial step (and they are logged in), the `/setup/password` and `/login` page are redirected to `/news`.

In the same sense as above, if a user has signed up and has completed the tutorial step (and they aren't logged in), the internal pages and the `/setup/password` are all redirected to `/login`.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -87,21 +87,23 @@ class Stringer < Sinatra::Base
     prebuild true unless ENV["RACK_ENV"] == "test"
   end
 
-  before do
-    I18n.locale = ENV["LOCALE"].blank? ? :en : ENV["LOCALE"].to_sym
+  def check_redirections
+    return unless needs_authentication?(request.path)
 
-    if !authenticated? && needs_authentication?(request.path)
-      session[:redirect_to] = request.path
-      redirect "/login"
+    if authenticated?
+      redirect to("/setup/tutorial") if needs_setup_complete?(request.path) && !UserRepository.setup_complete?
+    else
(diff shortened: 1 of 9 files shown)


gabrielpoca:
Hi @fearenales thank you for this contribution. I know that you have not gotten any response in a long time, and I'm sorry for that. I'm reading your changes and here are my thoughts:

* The code to redirect to one place or another introduces a lot of branching, and although you wrote good tests for it, I wonder how hard it will be to read and maintain. Some of the inline `ifs` hide the fact that we'll introduce three levels of conditions in some places. 
* What are your thoughts on using an authorization library, such as [pundit](https://github.com/elabs/pundit) for this behavior?

Please let me know what you think and if you're still interested in merging this code.

gabrielpoca:
Closing this PR for lack of activity.

Issue #422, opened 2016-02-29
Investigate upgrading to Feedjira 2.0


swanson:
https://github.com/feedjira/feedjira/tree/v2.0.0

Relevant code: 
https://github.com/swanson/stringer/blob/master/app/tasks/fetch_feed.rb
https://github.com/swanson/stringer/blob/master/app/utils/feed_discovery.rb

gabrielpoca:
This was solved in https://github.com/swanson/stringer/pull/448

Issue #380, opened 2015-07-30
Can't add some feeds


guiniol:
I'm not sure if this is related to my setup, or if others have experienced it.
Sometimes, when I try to add a feed, I get a `feed_not_found_error`, and after a few tries it goes through. I'm using stringer on a free instance of RedHat Cloud.
For example, I haven't been able to add this feed: https://copperhead.co/feed.xml

Tell me if there is any info I can provide. I'm not sure how to debug that.

Koronen:
To me this sounds like temporary issues with connectivity, so I'm not sure what we can do on our part. Does this affect fetching of any other feeds?

guiniol:
All my other feeds were refreshing properly. I tried the feed mentionned in my first comment for about a week without luck, so I posted the issue in case it was a problem bigger than connectivity.

Koronen:
Hey @guiniol, thank you for reporting this issue. I'm going to close this now due to inactivity. Feel free to re-open if you have any updates on this.

guiniol:
Alright. I haven't updated in a while, so I'll try and check how it goes.

matthiasballreich:
ticket should be re-openend, because this error still exists. Not only on the RSS in the first post, here: http://www.drwindows.de/external.php?do=rss&type=newcontent&sectionid=1&days=120&count=10, too :/ Stringer could not fetch the feed.

guiniol:
Definitely. I checked and forgot to come back and say that it still doesn't work.

Koronen:
None of these feeds pass the W3C Feed Validation Service at the time of this writing.

https://copperhead.co/feed.xml
> Server returned [Errno 1] _ssl.c:504: error:14077438:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert internal error [help]

http://www.drwindows.de/external.php?do=rss&type=newcontent&sectionid=1&days=120&count=10
> * description contains bad characters
> * description should not contain attachment140530" attribute
> etc.

matthiasballreich:
yeah i tried it myself with the W3C Feed Validator and the rist one is not valid, but the second one ist valid. W3C says it is a valid Feed. The Messages are only Recommendations like not using attachment attribute, etc.

Koronen:
The error with the `www.drwindows.de` feed appears to be related to Feedjira and the options we pass to it.
```
$ racksh
Rack::Shell v1.0.0 started in development environment.
[...]

PR #448, opened 2017-02-25
upgrade to feedjira 2.1.0 and upgrade to sinatra, sinatra-contrib 1.4.8

poligen:
Hello All:
I found this project last month and notice some open issues here. 
the #380, #422  are due to the upgrade [feedjira 2.0](http://feedjira.com/) .
The upgrade to above 2.0 needs only one parameter if you `fethc_and_parse`.
I try to remove options, and the remaining codes work fine.

I also notice there might be a problem if use ruby 2.3.0, and the console will print out `Thread.exclusive is deprecated, use Mutex`. And I found this sinatra-contrib [issue](https://github.com/sinatra/sinatra-contrib/issues/183). Therefore, I also upgrade to sinatra, and sinatra-contrib in Gemfile.


gabrielpoca:
Hi @poligen , thank you for this contribution. I would like to merge it, but I need to have the tests green before, could you fix them? The failing tests are related to the code you just changed.

poligen:
Hello @gabrielpoca , I also want to make the travis-ci to pass, but I look at the test code:

```ruby
 FeedDiscovery#discover returns false if url is not a feed and feed url cannot be discovered

     Failure/Error: feed = parser.fetch_and_parse(url)

     

       #<Double (anonymous)> received :fetch_and_parse with unexpected arguments

         expected: ("http://example.com", anything)

              got: ("http://example.com")
```

`fetch_and_parse with unexpected arguments`.

But in the feedjira 2.0, the fetch_and_parse only receive a single argument.

The `fetch_and_parse` source code in [feedjira/feed.rb](https://github.com/feedjira/feedjira/blob/a44637aeb29654ddd3ff1acc0856701918e29e6f/lib/feedjira/feed.rb)

```ruby
     def fetch_and_parse(url)
        response = connection(url).get
        unless response.success?
          raise FetchFailure, "Fetch failed - #{response.status}"
        end
        feed = parse response.body
        feed.feed_url = url
        feed.etag = response.headers['etag'].to_s.delete '"'

        feed.last_modified = parse_last_modified(response)
        feed
    end
```

And in feedjira README, it also suggests that `fetch_and_parse` only take one argument.

```markdown
## Fetching and Parsing

For many users, the `fetch_and_parse` method is what they use Feedjira for. This
url = "http://feedjira.com/blog/feed.xml"
feed = Feedjira::Feed.fetch_and_parse(url)
# => #<Feedjira::Parser::Atom...>
```

All the test code are related to this problem:
but I think in Feedjira 2.0, `fetch_and_parse` only take one argument

```ruby
rspec ./spec/utils/feed_discovery_spec.rb:15 # FeedDiscovery#discover returns false if url is not a feed and feed url cannot be discovered

rspec ./spec/utils/feed_discovery_spec.rb:24 # FeedDiscovery#discover returns a feed if the url provided is parsable

rspec ./spec/utils/feed_discovery_spec.rb:32 # FeedDiscovery#discover returns false if the discovered feed is not parsable

rspec ./spec/utils/feed_discovery_spec.rb:42 # FeedDiscovery#discover returns the feed if the discovered feed is parsable
```

Issue #427, opened 2016-03-22
Investigate alternatives to sinatra-assetpack?


Koronen:
```
remote:        Post-install message from sinatra-assetpack:
remote:        ##########################################################
remote:        # WARNING : sinatra-assetpack is not maintained anymore! #
remote:        ##########################################################
remote:        YOU MIGHT BE EXPOSED TO SECURITY ISSUES & BUGS.
remote:        See https://github.com/rstacruz/sinatra-assetpack/issues/197
```

pascalw:
Maybe could use https://github.com/hanami/assets? I'm not totally sure if you can use that without Hanami itself, but definitely worth looking into.

swanson:
I think Sprockets as better non-Rails support now, too.

gabrielpoca:
Solved in https://github.com/swanson/stringer/pull/449

PR #449, opened 2017-03-26
Replace sinatra-assetpack with sprockets

gabrielpoca:
In this change, I'm replacing Sinatra-assetpack with sprockets because assetpack is no longer maintained. Sprockets is in active development and has a very good non-Rails support.

I also noticed that we are not compiling assets on Heroku. For Rails applications, the buildpack will run `rake assets:precompile`, but there isn't anything similar for rack based applications. The best option would be fork the buildpack, but because each instance of String only serves one user, I don't think it will become a problem.

This change solves issue https://github.com/swanson/stringer/issues/427


gabrielpoca:
Hi @swanson , can I get your input on this?

swanson:
👍 Sprockets is fine with me

PR #454, opened 2017-08-06
Make Handling of Missing or Incorrect Api Key Consistent With Fever Api

jbrayton:
I am the developer of Unread, an RSS reader for iOS:

https://www.goldenhillsoftware.com/unread/

I recently implemented Fever URL validation. Unread performs this validation before asking for the customer's credentials. This validation attempts to confirm that the Fever URL is correct by issuing a request to the Fever API without an API key and verifying that the API returns the following JSON:

{ "api_version": 3, "auth": 0 }

This works with Fever, but it is not working with Stringer because Stringer responds with a 403 and no response body when the request does not contain the correct API key. This pull request makes Stringer consistent with Fever in this regard.

### DIFF
--- a/app/fever_api/response.rb
+++ b/app/fever_api/response.rb
@@ -15,6 +15,9 @@ require_relative "write_mark_feed"
 require_relative "write_mark_group"
 
 module FeverAPI
+
+  API_VERSION = 3
+
   class Response
     ACTIONS = [
       Authentication,
@@ -39,7 +42,7 @@ module FeverAPI
     end
 
     def to_json
-      base_response = { api_version: 3 }
+      base_response = { api_version: API_VERSION }
       ACTIONS
         .inject(base_response) { |a, e| a.merge!(e.new.call(@params)) }
         .to_json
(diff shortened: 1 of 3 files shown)


queuebit:
@jbrayton nicely done. I downloaded the new version of Unread after hearing the podcast and when I tried to login it to my stringer backend it failed and didn't have time to look into it at the time.

swanson:
This looks great -- one question: would it still work if we returned a 403, but with the body including the API version? It's unclear to be on if returning a 200 when the caller is unauthorized is mirroring the Fever API, but it feels slightly wrong semantically.

gglanzani:
I've "deployed" this version, and I can confirm that it's working now. 

@swanson I guess if that's consistent with the Fever API, we should keep it this way?

jbrayton:
@swanson Unread is looking for both the 200 response code that the Fever API returns as well as the JSON response body.

swanson:
Thanks -- agree with matching the Fever API so let's go with the 200 code.

PR #456, opened 2017-09-08
Add Dockerfile

pascalw:
This PR adds a Dockerfile so Stringer can easily be ran in Docker.

For convenience there's a single container running both Unicorn as well
as cronjobs.
This goes a little against the Docker mantra of running only a single
process per container, but hey :-)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,42 @@
+FROM ruby:2.3.3
+
+ENV RACK_ENV=production
+ENV PORT=8080
+
+EXPOSE 8080
+
+WORKDIR /app
+ADD Gemfile Gemfile.lock /app/
+RUN bundle install --deployment
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+      supervisor locales nodejs \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+RUN DEBIAN_FRONTEND=noninteractive dpkg-reconfigure locales \
+    && locale-gen C.UTF-8 \
+    && /usr/sbin/update-locale LANG=C.UTF-8
+
+ENV LC_ALL C.UTF-8
+
+ENV SUPERCRONIC_URL=https://github.com/aptible/supercronic/releases/download/v0.1.3/supercronic-linux-amd64 \
+    SUPERCRONIC=supercronic-linux-amd64 \
+    SUPERCRONIC_SHA1SUM=96960ba3207756bb01e6892c978264e5362e117e
+
+RUN curl -fsSLO "$SUPERCRONIC_URL" \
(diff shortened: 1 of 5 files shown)


swanson:
Yes, that would be great.

gabrielpoca:
@pascalw nice work 👍  I was thinking about writing a Dockerfile for a while. I'll give this a spin on the weekend.

pascalw:
@swanson @gabrielpoca documentation added, let me know what you think.

I've now included instructions to build the Docker image yourself. Once this is merged you could consider setting up an [Automated build to Docker Hub](https://docs.docker.com/docker-hub/builds/), so people can just `docker pull stringer` without having to build manually.

gabrielpoca:
@pascalw I managed to run it using the steps you described in the docs, although I think you're missing a step to create the docker network, something like `docker network create --driver bridge stringer`. Let me know if this makes sense or I am missing something.

I was thinking about the configuration for the reverse proxy, and even though it is helpful, that's no the only thing that you need to do to run it in production, and I wonder if it isn't enough to just leave the note saying that it is recommended to use a reverse proxy. What do you think?

pascalw:
Ah yes you're right @gabrielpoca, indeed I forgot instructions for the docket network, I'll add that.

About the reverse proxy; I think it's helpful to have ready-to-go configs that can be used to quickly get up and running. However what we might do is create a separate document that describes setting up a reverse proxy, because it applies to the [VPS setup](https://github.com/swanson/stringer/blob/master/docs/VPS.md). What do you think?

gabrielpoca:
@pascalw I agree with you. I think that what you have right now is ok 👍 would you mind looking into the tests?

pascalw:
@gabrielpoca just had a quick look at the tests and my guess is that Travis CI is now using Phantomjs 2.x but that's not supported [1] by the version of `mocha-phantomjs` used by this project. For me locally the mocha tests also freeze, both on this branch as well as master.

It's a little difficult to debug though as I can't verify if phantomjs 2.x is really the issue, since phantomjs 1.x doesn't seem to run on macOS Sierra and I have no other box to try this on. However I'll just try bumping the version of mocha-phantomjs in the travis config and see what happens.

[1] https://github.com/nathanboktae/mocha-phantomjs/issues/175

pascalw:
@gabrielpoca tests are green again. I've upgraded mocha-phantomjs and mocha. I'm going to move these commits to a separate PR as they are unrelated to this PR.

gabrielpoca:
@pascalw thank you much for taking the time! I've merged you PR.

pascalw:
I rebased this branch. Let me know how you want to proceed from here.

gabrielpoca:
I think it's ready to be merged. @swanson do you want to weigh in?

swanson:
Looks good to me – thanks for taking this on!

PR #457, opened 2017-09-09
Fix mocha tests

pascalw:
As discussed in #456 mocha tests are currently failing/hanging in Travis CI. I've also observed this behavior on my local machine.

Reason appears to be that the previously used version of `mocha-phantomjs` doesn't support phantomjs 2.x. It looks like Travis CI now uses phantomjs 2.x by default, hence these issues.

This PR upgrades `mocha-phantomjs` to latest version and also upgrades `mocha` to the latest version as the previously used version (not sure which version that was?) was not supported by this new version of `mocha-phantomjs`.

It's worth pointing out that this new version of `mocha-phantomjs` *supports* phantomjs 2.x, but includes version 1.9.7 which doesn't run on MacOS Sierra. To run the mocha specs on Sierra the following command can be used to run with a phantomjs version in your path: `mocha-phantomjs -p $(which phantomjs) http://localhost:4567/test`.


gabrielpoca:
Thank you @pascalw !

PR #460, opened 2017-09-14
Ignore un-existence of database.yml

maxmeyer:
* This makes unicorn happy if either DATABASE_URL is set or the config/database.yml is available
* Silence output of ActiveRecord - depending on LOG_LEVEL

### DIFF
--- a/config/unicorn.rb
+++ b/config/unicorn.rb
@@ -17,7 +17,11 @@ end
 after_fork do |_server, _worker|
   if defined?(ActiveRecord::Base)
     env = ENV["RACK_ENV"] || "development"
-    config = YAML.load(ERB.new(File.read("config/database.yml")).result)[env]
+    config = if ENV["DATABASE_URL"]
+               ENV["DATABASE_URL"]
+             else
+               YAML.load(ERB.new(File.read("config/database.yml")).result)[env]
+             end
     ActiveRecord::Base.establish_connection(config)
   end
 end
--- a/config/unicorn.rb
+++ b/config/unicorn.rb
@@ -23,5 +23,12 @@ after_fork do |_server, _worker|
                YAML.load(ERB.new(File.read("config/database.yml")).result)[env]
              end
     ActiveRecord::Base.establish_connection(config)
+
+    case ENV["LOG_LEVEL"]
+    when "info"
+      ActiveRecord::Base.logger.level = Logger::INFO
+    else
+      ActiveRecord::Base.logger.level = Logger::DEBUG
+    end
   end
 end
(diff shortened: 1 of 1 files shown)


gabrielpoca:
@maxmeyer thank you for the contribution. The code seems to be breaking a rule in rubocop, could you take care of it before I merge?

maxmeyer:
@gabrielpoca Yes, of course. Done.

maxmeyer:
@gabrielpoca Ok?

gabrielpoca:
Thank you @maxmeyer! I only now got the time to look at this change more carefully, if the `config/database.yml` already uses `<%= ENV["DATABASE_URL"] %>` in production, what is the scenario in which you need this change?

maxmeyer:
The problem is this: I removed the file in my container, as `active_record` can be used with `DATABASE_URL` defined, but starting stringer fails because `File.load` fails because the file does not exist.

gabrielpoca:
@maxmeyer I hope you don't mind me insisting, but it's not clear for me yet. If `config/database.yml` is already using `DATABASE_URL`, why are you removing the file?

maxmeyer:
@gabrielpoca Because I was unaware of this and overhead it in your comment.

Thinking more about it, the approach of having this indirection with `config/database.yml` seems to be less flexible as I'm not able to run a container in development environment configured via `DATABASE_URL`, correct?

gabrielpoca:
@maxmeyer yes, you are right, it's less flexible. It does work in production, but in development you would have to replace the file. If no one has complained about it not working when running in development, do you still think we should merge this?

maxmeyer:
@gabrielpoca I think yes, normally I try to build my docker images that they work in all environments the same. This https://12factor.net/config may give some more info why it makes sense to configure an application via environment variables.

@pascalw You added the docker support. What do you think? Shall the image work in all environments -
 prod, staging, ... the same?

pascalw:
@maxmeyer @gabrielpoca I do think ideally the Docker image works in every environment.

However I'm not convinced the changes in this PR are required to achieve that. For example I'm able to run Stringer with `RACK_ENV=development` and specifying a DB url like this:

```
docker run --rm -it -e DATABASE_URL="sqlite3:':memory:'" -e RACK_ENV=development -p 8080:8080 mdswanson/stringer
```

Afaik Activerecord prioritizes `DATABASE_URL` over `config.yml`.

bitboxer:
@mockdeep I think this one can be closed

mockdeep:
Sure, sounds good. @maxmeyer if you have any other thoughts on this, happy to hear them.

maxmeyer:
This is fine for me. I'm not using Stringer anymore.

PR #461, opened 2017-09-14
Make SECRET_TOKEN random by default

maxmeyer:
I found no place where the current default value of SECRET_TOKEN is used. So I propose to make it random by default using `SecureRandom.hex(32)`


gabrielpoca:
Thank you @maxmeyer 👍 I'm going to merge it

PR #463, opened 2017-09-23
Add docker-compose example and update docs

lubekpl:
If you're lazy like me, you try to automate your life. I've created this docker-compose example to quickly start up a Stringer instance and have all available options easily configured in one file, as opposed to running a containers manually using messy docker commands.


swanson:
Thanks!

PR #470, opened 2017-12-29
Strip url before parse it

LuYangP:
The link field of this [feed](https://www.qdaily.com/feed.xml) contains whitespace, got this error:
`bad URI(is not URI?):   (URI::InvalidURIError)`

Strip the url before parsing it solve the problem.


gabrielpoca:
Thank you 👍

PR #479, opened 2018-03-29
Prevents crash when both url and enclosure_url dont exist

naps62:
[this commit](https://github.com/swanson/stringer/commit/b4ae4c1d43c9f0bd597e76dcb3324ac6f89e8359) made the whole process fail when both url and enclosure_url do not exist. The reason is that `entry.enclosure_url.present?` throws an error, since the method doesn't exist at all

There's actually a much larger issue here, which is the error being silently hidden in `FetchFeed`, which just rescues every exception and flags the feed as red. This makes it very hard to debug errors

However I just fixed this one issue. I hope after re-deploying, my feeds will start working again, otherwise I'll have to get back at it

### DIFF
--- a/app/repositories/story_repository.rb
+++ b/app/repositories/story_repository.rb
@@ -82,7 +82,7 @@ class StoryRepository
   end
 
   def self.extract_url(entry, feed)
-    return entry.enclosure_url if entry.url.nil? && entry.enclosure_url.present?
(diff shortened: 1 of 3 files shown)


naps62:
FYI: just tested this on my running instance in Heroku. Previously, most feeds were not updating, due to a parsing error. After this fix, they started working (only one is still failing, I'll have to investigate that one later)

gabrielpoca:
I also tested locally 👍

PR #480, opened 2018-03-30
Fixes entry parse if there is no URL

naps62:
Parsing was failing if an entry had no defined URL. The parser was assuming it existed, and using it to expand all relative URLs

While it probably always exists for any well-made RSS feed, it won't be true if we're using something like [Kill The Newsletter](https://www.kill-the-newsletter.com/) to convert emails to RSS


gabrielpoca:
@naps62 could you squash before I merge it?

naps62:
@gabrielpoca done

PR #492, opened 2018-11-01
Sanitize feed titles

### LINKED ISSUE #490: Script tags not stripped from feed titles on /feeds endpoint
I believe I have discovered an XSS issue in Stringer (the most recent version pushed to Docker Hub).

Here's a POC for it:

1. Run Stringer using the instructions in https://github.com/swanson/stringer/blob/master/docs/docker.md (The SQLite and Docker Compose instructions instructions both work.)
2. Serve an RSS feed that has an entity-encoded script tag in its title. (i.e. \&lt;script\&gt;alert('xss')\&lt;/script\&gt; )
3. Add your webserver for that feed to Stringer
4. Go to the /feeds endpoint in your browser. You should see the alert.

I have created an example RSS feed that demonstrates this at https://gist.github.com/rwhogg/1a7b9e9f82db6fdfb266af2cd09cfb9a

Koronen:
Attempt to fix a recently reported XSS vulnerability.

We reuse the same sanitization strategy as for _story_ titles.

Also add a `StoryRepository` test for title-based XSS for good measure.

Fixes #490.

### DIFF
--- a/app/commands/feeds/add_new_feed.rb
+++ b/app/commands/feeds/add_new_feed.rb
@@ -1,4 +1,5 @@
 require_relative "../../models/feed"
+require_relative "../../utils/content_sanitizer"
 require_relative "../../utils/feed_discovery"
 
 class AddNewFeed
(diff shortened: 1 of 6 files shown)


swanson:
LGTM 👍

PR #516, opened 2020-05-13
Update command for setting APP_URL on Heroku

mattreduce:
* `web-url` in the `heroku apps:info` output is now `web_url`
* Command was returning an empty string and setting that as the value of `APP_URL`


swanson:
Thanks!

PR #533, opened 2021-02-22
Update to activerecord 5

dperson:
Followed the instructions at https://www.fastruby.io/blog/rails/upgrades/upgrade-rails-from-4-2-to-5-0.html to migrate to the newer version. Then fixed the `pec/support/active_record.rb` calls to work with the new API.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,7 @@ group :development, :test do
   gem "timecop", "~> 0.8"
 end
 
-gem "activerecord", "~> 4.2.6"
+gem "activerecord", "~> 5.0"
 gem "bcrypt", "~> 3.1"
 gem "delayed_job", "~> 4.1"
 gem "delayed_job_active_record", "~> 4.1"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,20 +1,19 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    activemodel (4.2.10)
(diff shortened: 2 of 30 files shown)


mockdeep:
@dperson Thanks for jumping into this!

One thing that concerns me about the upgrade is [`belongs_to` being required by default](https://github.com/rails/rails/pull/18937/files). In my short time working on Stringer, it seems like things aren't very tightly constrained, so I could see that throwing an error on any missing data. We might want to set that to `false` for the time being:

```rb
config.active_record.belongs_to_required_by_default = false
```

dperson:
@mockdeep it has been running well for me, but I'll be glad to change the configuration to maintain the historical behavior.

mockdeep:
@dperson actually, it looks like the configuration defaults to `false` outside of Rails apps. If it was enabled, I'm pretty sure we'd have failing tests. New feeds by default don't have a group, so it wouldn't allow you to add a new feed if it was required. You can remove the `config/application.rb` you added. (I don't think it's doing anything anyway, since the file doesn't seem to be getting loaded anywhere.)

Sorry for the back and forth!

dperson:
@mockdeep no problem. I deleted the commit that added the file, and pushed it back to the branch. I'm just glad that there are fixes / updates coming once again. Thank you!

mockdeep:
> I'm just glad that there are fixes / updates coming once again.

Me too. Thanks!

PR #472, opened 2018-01-24
Move to using Feedjira 3.0

mikeastock:
Hello 👋! I'm a maintainer on Feedjira and are hoping you guys are open to potentially testing out and giving us feedback on the new interface in Feedjira 3.0.

Feedjira removed it's built-in HTTP fetching so that has been replaced with HTTParty. The underlying parsing code has not changed and Feedjira is in a beta status while we try to nail down the new interface.

We are open to any and all feedback about Feedjira. Thanks!

PR #535, opened 2021-02-23
update to feedjira 3

dperson:
This is a functional and tested version of #472
The 2.x version HEAD is on has been EOL since 2018.


Gemfile
```diff
@@ -30,7 +30,8 @@ gem "bcrypt", "~> 3.1"
 gem "delayed_job", "~> 4.1"
 gem "delayed_job_active_record", "~> 4.1"
 gem "feedbag", "~> 0.9.5"
-gem "feedjira", "~> 2.1.3"
+gem "feedjira", "~> 3.0"
+gem "httparty"
```
mockdeep, line 34: It's not a huge deal, but I wonder if we'd be better off using Ruby's built in `Net::HTTP` library rather than adding a new dependency. I know `Net::HTTP` doesn't have the greatest reputation for usability, but we're only making a couple of requests so it doesn't seem like it'll make a huge difference. I'll leave it up to you, but my bias is towards minimizing dependencies unless they save us a ton of time and headache.

dperson:
Okay, I had just gone with the suggested solution from the `feedjira` documentation and examples, but that is a completely valid point. I've made the changes and fixed the CI tests.

Let me push the change to my personal instance and make sure it doesn't break horribly.

dperson:
Hmm, I think that it's not getting all of my feeds any longer (with Net::HTTP). There are a few that I normally get updates every morning from that I haven't seen, and I have a lot more amber and red feeds in the feed viewer.

My experience with Ruby (besides this project) was writing some extensions for Puppet to pull in information from internal corporate sources for configuring hosts... I really don't know where to go to try and debug / fix this issue.

Unless you have some pointers, I'm inclined to go back to their suggested HTTP client...

dperson:
Yeah after reverting to HTTParty my feeds that I wasn't seeing updates on all came in. I don't know why it didn't work with Net::HTTP.

mockdeep:
@dperson are you able to share one or two of the feeds that didn't work for you? I'll see if I can figure out the logging situation and get a better picture of what is going wrong. Could be we're better off with HTTParty, but I'd like to understand what is happening first.

dperson:
Two that I can remember coming in immediately after swapping back to HTTParty are:

http://www.quotationspage.com/data/qotd.rss
http://www.sffaudio.com/?feed=rss2

Thanks!

mockdeep:
Okay, so I'm poking around with it and for some reason `Net::HTTP.get` returns an empty string for the second feed, but `HTTParty.get` returns an XML document as expected. That's just weird behavior, so I guess we'd better stick with `HTTParty`. Sorry, another flip flop!

I'm not sure if it's a Feedjira thing or a Net::HTTP thing, but I wasn't able to add your qotd feed on your branch. I am able to add it on `master`, though. If I click the link it redirects me to a Feedburner page which I am able to add as a feed on your branch.

mockdeep:
Also, maybe you're already aware of it, but you can test these things out locally so you don't risk messing up your production feeds. I use `foreman`:

```sh
$ gem install foreman
$ foreman start
```

You can run background jobs with `rake work_jobs` and fetch feeds manually with `rake fetch_feeds`.

dperson:
Yeah, I really don't know why Net::HTTP isn't reliable for queries... actually I just thought of something, I wonder if it doesn't follow redirects? If the formerly build into feedjira and HTTParty do... of course I just finished reverting back to HTTParty. How about we merge it, and I'll work at getting Net::HTTP to follow redirects?

dperson:
Yeah, Net::HTTP doesn't follow redirects automatically, you have to check the status and manually follow them, yuck.

mockdeep:
Ah, wow. Great sleuthing! So there is one other core ruby alternative, which is OpenURI. It appears that that properly follows redirects:

```rb
require "open-uri"

URI("http://www.quotationspage.com/data/qotd.rss").read
# => "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\r\n<?xml-stylesheet type=\"text/xsl\" ...
```

So it looks like we could use that instead. But I'll leave it up to you. HTTParty seems like it's probably more robust.

dperson:
If it's okay with you, I'd like to merge the HTTParty version, and I'll work on redoing it with OpenURI later. That way if it is broken somehow, we can roll back to the version upstream recommends / tests.

mockdeep:
Sounds good to me!

PR #539, opened 2021-02-25
CI: freeze bundler lockfile

mockdeep:
This should make it so that no gem updates get merged without
corresponding `Gemfile.lock` updates.


mockdeep:
Closing for now, I want to do some additional checking to make sure we're doing retries properly on TravisCI.

PR #540, opened 2021-02-26
update gemfile dependencies, and fix new rubocop warnings

dperson:
This is primarily to update stringer dependencies and the ruby version, but that updated `rubocop` (ruby linter) and `capybara` (used in testing). With the new `rubocop` version, a number of small changes were prompted to fix style issues, disable notifications in a couple places, remove a disable because it's no longer needed, etc. With the new version of `capybara` a configuration setting was needed, which required a couple small changes.


mockdeep:
@dperson this many upgrades at the same time makes me a tad nervous. Would you be up for breaking it up into a handful of PRs? It might be nice to have the Rubocop changes on their own PR, and I generally like to give a Ruby upgrade its own PR as well.

dperson:
@mockdeep That's fine by me. I had made most of these changes over time in my branch and been running them for quite a while. I split off the ones that seemed to have the most core use in the platform rack & sinatra, activerecord, and feedjira. To me these were the long tail support gems. I'm more than happy to break them up into smaller chunks though. Do you have a preference on what order? and did you want just the 3 mentioned?

* rubocop
* remainder
* ruby

OR based on Gemfile groups:

* dev gems
* dev test gems
* production (pg (postgres) and unicorn)
* remainder
* ruby

**NOTE**: I'm working on the OpenURI conversion of feedjira, and hope to have that for you soon too. Working at figuring out the test system, it's not happy with the new `read` function being called.

mockdeep:
@dperson good question. I like your idea of splitting off by gem group, though I don't think they need to be split up too much. Maybe something like:

* rubocop
* other dev/test gems
* production gems
* ruby

That's assuming you aren't doing any major upgrades in the production gems, yet. For something like ActiveRecord I might bump it individually to 5.x and then 6.x.

For the OpenURI stuff, feel free to push up a branch and I can take a look. Testing is one of my specialties.

dperson:
Sounds good to me.

PR #543, opened 2021-02-27
update test gems and dependencies

dperson:
fix a rubocop alert from the newer version faker
work around a capybara change


Gemfile
```diff
@@ -13,16 +13,16 @@ group :development do
...
-  gem "capybara", "~> 2.6"
-  gem "coveralls", "~> 0.7", require: false
-  gem "faker", "~> 1.2"
-  gem "pry-byebug", "~> 1.2"
-  gem "rack-test", "~> 1.1"
-  gem "rspec", "~> 3.4"
-  gem "rspec-html-matchers", "~> 0.7"
-  gem "shotgun", "~> 0.9"
+  gem "capybara"
+  gem "coveralls", "> 0.8", require: false
```
mockdeep, line 17: Can we not remove the version constraint here, too?

Gemfile.lock
```diff
@@ -12,42 +12,40 @@ GEM
...
-      thor (~> 0.19.1)
-      tins (~> 1.6.0)
+      thor (>= 0.19.4, < 2.0)
+      tins (~> 1.6)
     crass (1.0.5)
-    debugger-linecache (1.2.0)
     delayed_job (4.1.8)
       activesupport (>= 3.0, < 6.1)
     delayed_job_active_record (4.1.0)
       activerecord (>= 3.0, < 5)
```
mockdeep, line 42: @dperson have you by any chance been making manual edits to `Gemfile.lock` or selectively committing lines? There's some weird stuff happening when I try to add/change gems. Based on this line, it looks like you shouldn't have been able to update `activerecord` to `5.2`.

Gemfile
dperson: No, because there is some recoverable conflict in dependencies. If it's not specified to be greater than 0.8 it installs a 0.7.2 (if I recall correctly) from 2014, and it doesn't support all the features implemented in the code base.

It's selected (or so I believe) because that was the last version that doesn't have a dependency on an older version of another gem we use.

Gemfile.lock
dperson: The final version (from the big update you wanted split up) was updated by `bundle lock --update` but I was trying to do these updates by hand to update just the ones effected by this change. Did I fat finger something?

mockdeep: It looks like #533 should not have been possible without also updating `delayed_job_active_record`. I'm a little surprised bundler let it happen. If you run `bundle update --group=test` on `master` it will throw an error. #546 should fix it.

It's not a huge deal, but in general, I would avoid manually updating `Gemfile.lock` or selectively adding lines via something like `git add -p`. Instead, it's best to let bundler resolve any updates to make sure everything is in a consistent state. I'm not familiar with `bundle lock --update`, but when I'm updating gems I'll general use some variation on [`bundle update`](https://bundler.io/man/bundle-update.1.html). You can use `bundle update group=test` or `bundle update --minor` or `bundle update activerecord`.

mockdeep: Sorry, looks like you'll need to rebase this one.

Gemfile.lock
```diff
@@ -156,22 +155,33 @@ GEM
...
+||||||| parent of a3d2b33 (update test gems and dependencies)
+    timecop (0.8.0)
+    tins (1.6.0)
+    tzinfo (1.2.5)
+=======
+    timecop (0.9.4)
+    tins (1.28.0)
+      sync
+    tzinfo (1.2.5)
+>>>>>>> a3d2b33 (update test gems and dependencies)
```
mockdeep, line 184: Looks like some git conflicts snuck in here. For this, I'd probably:

```
git checkout master Gemfile.lock
bundle update --group=test
```

Gemfile.lock
dperson: Ah, sorry about that. I'm not a ruby expert, and appreciate the pointer on `bundle`.

The rebase is complete. Thanks!

mockdeep: Yeah, this kind of shakes my faith in Bundler. I poked around for a while but wasn't able to find a good way to get it to throw an error, aside from trying to add a new gem.

PR #565, opened 2021-04-11
DB: update schema.rb

mockdeep:
This is auto-generated when I run `rake db:migrate`. It looks like the
formatting just changed some in later versions of ActiveRecord.


mockdeep:
Just handling it in [the AR upgrade](https://github.com/swanson/stringer/pull/566), since there are more changes to the `schema.rb` file in 6.1.

PR #570, opened 2021-05-01
Deps: switch from Unicorn to Puma

mockdeep:
**What**

This switches the server we use from Unicorn to Puma

**Why**

* Puma is the blessed option [on Heroku][heroku].
* It [works with Capybara][capybara] out of the box.
* Ruby 3 [removes `webrick`][ruby3] from the core libraries, so we would
  need to add it as a dependency for Capybara when upgrading. Might as
  well consolidate on a single server.

**Notes**

I also moved `pg` out of the `production` group.

[heroku]: https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#recommended-default-puma-process-and-thread-configuration
[capybara]: https://github.com/teamcapybara/capybara#drivers
[ruby3]: https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -33,6 +33,9 @@ Style/Documentation:
 Style/DoubleNegation:
   Enabled: false
 
+Style/MissingElse:
+  Enabled: false
+
 Style/NumericLiterals:
   Enabled: false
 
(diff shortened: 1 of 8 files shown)


mockdeep:
Pushing up the PR, but I want to run this on Heroku for a couple of weeks before merging.

mockdeep:
I've been testing this for the better part of a year and it hasn't had any issues.

PR #571, opened 2021-05-18
update to match swanson/stringer

geemus:


geemus:
Oops, trying to update my copy and apparently got things backwards :(.

PR #574, opened 2021-06-05
Fix the building of docker image

### LINKED ISSUE #523: Update Docker image
The Docker image [here](https://hub.docker.com/r/mdswanson/stringer) seems be outdated, comparing to the `master` branch.

Any chance to update it? Seems like auto-build is not working as expected (from #458).

vitkabele:
Fixes #523.
Building Docker image was failing due to the outdated ruby base image. Updated to ruby:latest.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.3.3
+FROM ruby:2.7.2
 
 ENV RACK_ENV=production
 ENV PORT=8080
@@ -7,7 +7,7 @@ EXPOSE 8080
 
 WORKDIR /app
 ADD Gemfile Gemfile.lock /app/
-RUN bundle install --deployment
+RUN bundle config set --local deployment 'true' && bundle install
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
@@ -36,7 +36,7 @@ ADD docker/start.sh /app/
 ADD . /app
 
 RUN useradd -m stringer
-RUN chown -R stringer:stringer /app
+RUN chown -R stringer:stringer /app /etc/supervisord.conf
 USER stringer
 
 CMD /app/start.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ EXPOSE 8080
 
 WORKDIR /app
 ADD Gemfile Gemfile.lock /app/
-RUN bundle config set --local deployment 'true' && bundle install
+RUN bundle config set --local deployment true && bundle install
 
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


vitkabele:
This should also fix the automated container build.

Dockerfile
```diff
@@ -7,7 +7,7 @@ EXPOSE 8080
 
 WORKDIR /app
 ADD Gemfile Gemfile.lock /app/
-RUN bundle install --deployment
+RUN bundle config set --local deployment 'true' && bundle install
```
mockdeep, line 10: I don't think we need the quotes around `'true'` here. All CLI arguments get passed through as strings to Ruby.

Dockerfile
```diff
@@ -36,7 +36,7 @@ ADD docker/start.sh /app/
 ADD . /app
 
 RUN useradd -m stringer
-RUN chown -R stringer:stringer /app
+RUN chown -R stringer:stringer /app /etc/supervisord.conf
```
mockdeep, line 39: Can you tell me more about why this was necessary?

mockdeep (commented):
Thanks so much for this @vitkabele! A couple of minor comments, but looks pretty good!

Dockerfile
vitkabele: Otherwise the supervisord is failing with error "Cannot read /etc/supervisord.conf"

Dockerfile
vitkabele: Yes, it could. This is what the bundler itself told me that is the preferred way :).

vitkabele:
Sadly all attempts to run the container ends with: 
```

2021-06-06 10:48:39,502 DEBG 'unicorn' stdout output:

2021-06-06 10:48:39 - ActiveRecord::RecordNotFound - Couldn't find User with 'id'=1:

	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/core.rb:155:in `find'
	/app/app/repositories/user_repository.rb:7:in `fetch'
	/app/app/helpers/authentication_helpers.rb:20:in `current_user'
	app/views/layout.erb:22:in `block in singleton class'
	app/views/layout.erb:-5:in `instance_eval'
	app/views/layout.erb:-5:in `singleton class'
	app/views/layout.erb:-7:in `__tilt_14856460'
	...
	/usr/local/bin/bundle:22:in `load'
	/usr/local/bin/bundle:22:in `<main>'
```

Any ideas? The problem is clearly with initial user creation, but I don't have time to solve it now. This is WIP until resolved.

vitkabele:
Should work now. Please check

PR #614, opened 2022-03-29
Replacing removed update_attributes method

### LINKED ISSUE #612: Mark as read broken?
This weekend I upgraded my 2 year old stringer to the latest code, sadly this broke the "marked as read" in my reeder apps somehow. Not sure what happend there. Anyone else having the same issue?

bitboxer:
This will replace the method `update_attributes` that was removed in ActiveRecord 6.1. 

Fixes #612

### DIFF
--- a/app/commands/stories/mark_as_read.rb
+++ b/app/commands/stories/mark_as_read.rb
@@ -7,6 +7,6 @@ class MarkAsRead
   end
 
   def mark_as_read
-    @repo.fetch(@story_id).update_attributes(is_read: true)
+    @repo.fetch(@story_id).update(is_read: true)
   end
 end
--- a/app/commands/stories/mark_as_read.rb
+++ b/app/commands/stories/mark_as_read.rb
@@ -1,12 +1,11 @@
 require_relative "../../repositories/story_repository"
 
 class MarkAsRead
-  def initialize(story_id, repository = StoryRepository)
+  def initialize(story_id)
     @story_id = story_id
-    @repo = repository
   end
 
   def mark_as_read
-    @repo.fetch(@story_id).update(is_read: true)
+    StoryRepository.fetch(@story_id).update!(is_read: true)
   end
 end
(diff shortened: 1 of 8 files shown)


app/commands/stories/mark_as_read.rb
```diff
@@ -7,6 +7,6 @@ def initialize(story_id, repository = StoryRepository)
   end
 
   def mark_as_read
-    @repo.fetch(@story_id).update_attributes(is_read: true)
+    @repo.fetch(@story_id).update(is_read: true)
```
mockdeep, line 10: I think I might use `update!` for these so that it will fail loudly when it doesn't work for some reason.

spec/commands/stories/mark_as_read_spec.rb
```diff
@@ -9,7 +9,7 @@
 
     it "marks a story as read" do
       command = MarkAsRead.new(1, repo)
-      expect(story).to receive(:update_attributes).with(is_read: true)
+      expect(story).to receive(:update).with(is_read: true)
```
mockdeep, line 12: This is one reason why I generally avoid stubbing methods. Does it work if we change it to something like:

```rb
expect { command.mark_as_read }.to change(story, :is_read).to(true)
```

mockdeep (commented):
Thanks so much for fixing this!

spec/commands/stories/mark_as_read_spec.rb
bitboxer: I would prefer that, too. But if you open the file you see that everything is stubbed out. I will change all files to actual model instances that hit the db.

app/commands/stories/mark_as_read.rb
bitboxer: Yes, good idea.

bitboxer:
@mockdeep I refactored the tests

mockdeep:
Excellent, thank you!

PR #686, opened 2022-12-16
Security: always enforce SSL

mockdeep:
I had to add a monkey patch for `Rack::Test` as they don't seem to have
a good way to enable SSL by default otherwise.


mockdeep:
This breaks dev mode. I'll need to rethink my approach to this.

PR #761, opened 2022-12-31
Dev: remove racksh

mockdeep:
It currently [doesn't work with Ruby 3.2][r3]. For our purposes, we can
instead replace it with a small rake task.

[r3]: https://github.com/sickill/racksh/pull/15


dentarg:
I forked racksh and made it work with Ruby 3.2: https://github.com/spinels/racksh

It is published as `spinels-racksh` (https://rubygems.org/gems/spinels-racksh)

mockdeep:
@dentarg thanks. Stringer is now on Rails, so we can use the built in `rails console`.

PR #849, opened 2023-01-29
Fixed routing for reeder app

bitboxer:
Reeder app assumes that the path ends on a `/`, e.g. `http://www.example.com/fever/`. It always adds that `/`.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,8 +13,8 @@ class Stringer < Sinatra::Base
   end
 
   match("/", to: "stories#index", via: :get)
-  match("/fever", to: "fever#index", via: :get)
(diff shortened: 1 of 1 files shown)


mockdeep:
Thank you!

mockdeep:
@bitboxer out of curiosity, how is the Stringer fever api working out for you? I've been touching on it in passing and it doesn't look very complete to me.

bitboxer:
It works perfectly in my reeder app for years now. Reeder marked the fever integration as „deprecated“ and recommends the google reader api. So might be that he will remove it in the future. If he does I need to investigate in what parts of the google reader api it’s using and adding that to stringer. I am never using stringer directly besides when managing my subscripted feeds.

PR #1053, opened 2023-07-06
adjust supercronic download to allow multiarch builds

guidopetri:
The download of `supercronic` in the Dockerfile hardcodes `amd64` as the architecture, but we can make it flexible to allow for multiarch builds.

For this purpose I added all three of the SHA1SUMs and just select the "correct" one to use based on the `ARCH` build argument (which has a default of `amd64`).

This built successfully for me with the command:
```
docker build -t stringer-arch-test --build-arg ARCH=arm64 .
```
(as well as `arm` and `amd64`)

I changed the shell that executes the Dockerfile to bash to take advantage of the `${!ENV_VAR_CONTAINING_OTHER_ENV_VAR_NAME}` construction - the `SUPERCRONIC_SHA1SUM` arg contains one of `SUPERCRONIC_{amd64,arm,arm64}_SHA1SUM`, and we need to evaluate this value as an env var. The easiest way to do this with `/bin/sh` seems to be `eval echo \$$SUPERCRONIC_SHA1SUM`, but that's a bit unsafe since we're using `eval` so I selected to switch to bash instead.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,12 +22,18 @@ RUN DEBIAN_FRONTEND=noninteractive dpkg-reconfigure locales \
 
 ENV LC_ALL C.UTF-8
 
-ENV SUPERCRONIC_URL=https://github.com/aptible/supercronic/releases/download/v0.1.3/supercronic-linux-amd64 \
-    SUPERCRONIC=supercronic-linux-amd64 \
-    SUPERCRONIC_SHA1SUM=96960ba3207756bb01e6892c978264e5362e117e
+ARG ARCH=amd64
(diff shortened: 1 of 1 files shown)


guidopetri:
@mockdeep / @hozza I think this is ready for review :)

mockdeep (approved):

mockdeep:
This all looks pretty reasonable to me. Since we've moved to GoodJob, we might eventually be able to use their [repeating jobs support](https://github.com/bensheldon/good_job#cron-style-repeatingrecurring-jobs), but I think it makes sense to get this working for now.

@hozza do you want to take a look before I merge this?

mockdeep:
thanks @guidopetri!

PR #1065, opened 2023-07-22
update circleci to build image and push to dockerhub, pending approval

guidopetri:
Confidence that this will work: 0% :)

I renamed the previous `build` job to `run_tests`, since it seemed more apt, if we're adding a "build the docker image" step. I think this requires a change on the github repo / PR config, though, so I'm fine reverting that change.

This PR hopefully adds the `build-and-publish-image` workflow, which waits for tests to run + waits for an approval from a maintainer / someone with merge access to the repo to run it. I think we need to create the `stringer-rss` Dockerhub user/organization, and we must use `DOCKER_LOGIN` and `DOCKER_PASSWORD` env vars but I'm not sure where to place them in the CI file in a safe way. (is this something we set up in CircleCI itself as a secret?)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,18 +109,17 @@ jobs:
           command: |
             COVERAGE=false bundle exec rspec spec/system/
 
-
 workflows:
   test-build-and-publish-image:
     jobs:
       - run-tests
-      - wait-for-approval:
-          type: approval
       - docker/publish:
-          cache_from: $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME:latest
-          image: $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
-          tag: latest
+          cache_from: stringerrss/stringer:latest
+          image: stringerrss/stringer
+          tag: latest,$CIRCLE_SHA1
           update-description: false
           requires:
-            - wait-for-approval
             - run-tests
+          filters:
+            branches:
+              only: main
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,9 +3,10 @@ version: 2.1
 orbs:
   browser-tools: circleci/browser-tools@1.2.3
   node: circleci/node@5.0.0
+  docker: circleci/docker@2.2.0
 
 jobs:
-  build:
+  run-tests:
     parallelism: 1
     docker:
       - image: cimg/ruby:3.2.2-browsers
@@ -107,3 +108,19 @@ jobs:
           name: Run system tests
           command: |
             COVERAGE=false bundle exec rspec spec/system/
+
+
+workflows:
+  test-build-and-publish-image:
+    jobs:
+      - run-tests
+      - wait-for-approval:
+          type: approval
+      - docker/publish:
+          cache_from: $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME:latest
+          image: $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
+          tag: latest
+          update-description: false
+          requires:
+            - wait-for-approval
+            - run-tests


guidopetri:
@mockdeep action item requests for you:

- (decide whether we want to use Dockerhub or GHCR)
- create the `stringer-rss` org on Dockerhub
- test the "wait for approval" step on CircleCI - you should be able to just click somewhere to unblock the build
- (potentially) change the github PR config to require `run-tests` instead of `build`
- and finally, review that I haven't screwed up anything on the CircleCI config file :)

.circleci/config.yml
```diff
@@ -107,3 +108,19 @@ jobs:
...
+      - wait-for-approval:
+          type: approval
+      - docker/publish:
+          cache_from: $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME:latest
+          image: $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
+          tag: latest
+          update-description: false
+          requires:
+            - wait-for-approval
+            - run-tests
```
mockdeep, line 126: Is `wait-for-approval` needed here? Unless there's a strong reason for it, maybe we can publish on any merge to `main`. There's a `filters` key that allows us to run a workflow only on a particular branch:

```suggestion
          requires:
            - run-tests
          filters:
            branches:
              only: main
```

.circleci/config.yml
```diff
@@ -107,3 +108,19 @@ jobs:
...
+workflows:
+  test-build-and-publish-image:
+    jobs:
+      - run-tests
+      - wait-for-approval:
+          type: approval
+      - docker/publish:
+          cache_from: $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME:latest
+          image: $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
+          tag: latest
```
mockdeep, line 122: Should this also have a commit tag so it's easy to reference a particular version? Not sure if that happens automatically.

```suggestion
          tag: latest,$CIRCLE_SHA1
```

.circleci/config.yml
```diff
@@ -107,3 +108,19 @@ jobs:
...
+
+workflows:
+  test-build-and-publish-image:
+    jobs:
+      - run-tests
+      - wait-for-approval:
+          type: approval
+      - docker/publish:
+          cache_from: $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME:latest
+          image: $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
```
mockdeep, line 121: Dockerhub doesn't allow `-` in usernames, so [I signed up as the user `stringerrss`](https://hub.docker.com/repository/docker/stringerrss/stringer). I could set it as an env variable on CircleCI, or we could hard code it here:

```suggestion
          cache_from: stringerrss/stringer:latest
          image: stringerrss/stringer
```

mockdeep (commented):
Thanks so much for this @guidopetri! I've added some suggestions. I added the env variables on CircleCI, but [the build still didn't work](https://app.circleci.com/pipelines/github/stringer-rss/stringer/985/workflows/395494a1-b5ce-48a8-b65e-b957c001c40f/jobs/988). I think that's because CircleCI doesn't pass env variables to forked builds for security, so I think that probably makes my suggestion to build only on `main` a better way.

.circleci/config.yml
guidopetri: No, it wasn't needed - I thought you might want to control when the build/publish happens. It only triggering on `main` makes sense to me!

.circleci/config.yml
guidopetri: Yep, we can do that. I couldn't figure out yesterday how to get the SHA, hehe.

.circleci/config.yml
guidopetri: The example docs had these env vars within - I thought it was one of those things that it automatically provides certain env vars. I can hardcode it in :)

guidopetri:
> I think that's because CircleCI doesn't pass env variables to forked builds for security, so I think that probably makes my suggestion to build only on main a better way.

How confident are you that it'll work? :D We can try merging to `main` and seeing what happens, and opening another PR for fixes if necessary.

As for the commit I just added, I thought it'd be more convenient for you to review a separate commit, but we can squash everything later.

.circleci/config.yml
mockdeep: Yeah, I think CircleCI sets those env variables based on the Github org/repo, so we'd need to do something else because of the `-` in `stringer-rss`.

mockdeep:
Looks great! Thanks so much for this!

> How confident are you that it'll work?

Well, I'm more confident that we can revert if it doesn't work out, so worth a try!

PR #1071, opened 2023-08-06
Update Docker-compose file

Ni-g-3l:
Hey ! 

I was in lot of trouble on deploying the stringer docker-compose to my Portainer. So this is a little PR to fix it.

Feel free to report any mistakes in the PR (I'm at the beginning of my Open Source adventure). 

* Update postgres alpine version
* Update README and Docker Docs
* Update docker-compose file

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,27 +1,40 @@
-version: '2'
+version: '3.8'
+
 services:
   postgres:
-    image: postgres:9.5-alpine
+    image: postgres:12-alpine
+    hostname: stringer-postgres
     restart: always
+    networks:
+      - stringer
     volumes:
-      - ~/stringer:/var/lib/postgresql/data
+      - ./postgres:/var/lib/postgresql/data
     environment:
-      - POSTGRES_PASSWORD=super_secret_password
+      - POSTGRES_PASSWORD=<PLEASE_ENTER_YOUR_PASSWORD>
       - POSTGRES_USER=db_user
       - POSTGRES_DB=stringer
 
   web:
-    image: mockdeep/stringer
+    image: mockdeep/stringer:latest
     build: .
     depends_on:
       - postgres
     restart: always
     ports:
-      - 80:8080
+      - '127.0.0.1:8080:8080'
+    networks:
+      - stringer-network
     environment:
-      - SECRET_KEY_BASE=<your configuration>
-      - ENCRYPTION_PRIMARY_KEY<your configuration>
-      - ENCRYPTION_DETERMINISTIC_KEY=<your configuration>
-      - ENCRYPTION_KEY_DERIVATION_SALT=<your configuration>
+      - SECRET_KEY_BASE=$(openssl rand -hex 64)
+      - ENCRYPTION_PRIMARY_KEY=$(openssl rand -hex 64)
+      - ENCRYPTION_DETERMINISTIC_KEY=$(openssl rand -hex 64)
+      - ENCRYPTION_KEY_DERIVATION_SALT=$(openssl rand -hex 64)
+      - FETCH_FEEDS_CRON="*/5 * * * *" # optional
+      - CLEANUP_CRON="0 0 * * *" # optional
       - PORT=8080
-      - DATABASE_URL=postgres://db_user:super_secret_password@postgres:5432/stringer
+      - DATABASE_URL=postgres://db_user:<PLEASE_ENTER_YOUR_PASSWORD>@stringer-postgres:5432/stringer
+
+networks:
+  stringer:
+    external: false
+    name: stringer
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3.8'
 
 services:
   postgres:
-    image: postgres:12-alpine
+    image: postgres:15-alpine
     hostname: stringer-postgres
     restart: always
     networks:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,10 +25,10 @@ services:
     networks:
       - stringer-network
     environment:
-      - SECRET_KEY_BASE=$(openssl rand -hex 64)
-      - ENCRYPTION_PRIMARY_KEY=$(openssl rand -hex 64)
-      - ENCRYPTION_DETERMINISTIC_KEY=$(openssl rand -hex 64)
-      - ENCRYPTION_KEY_DERIVATION_SALT=$(openssl rand -hex 64)
+      - SECRET_KEY_BASE=<your configuration>
+      - ENCRYPTION_PRIMARY_KEY<your configuration>
+      - ENCRYPTION_DETERMINISTIC_KEY=<your configuration>
+      - ENCRYPTION_KEY_DERIVATION_SALT=<your configuration>
       # - FETCH_FEEDS_CRON="*/5 * * * *" # optional
       # - CLEANUP_CRON="0 0 * * *" # optional
       - PORT=8080
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     networks:
       - stringer
     volumes:
-      - ./postgres:/var/lib/postgresql/data
+      - ~/stringer/postgres:/var/lib/postgresql/data
     environment:
       - POSTGRES_PASSWORD=<PLEASE_ENTER_YOUR_PASSWORD>
       - POSTGRES_USER=db_user
@@ -21,7 +21,7 @@ services:
       - postgres
     restart: always
     ports:
-      - '127.0.0.1:8080:8080'
+      - '8080:8080'
     networks:
       - stringer-network
     environment:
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -1,27 +1,40 @@
...
+      - ./postgres:/var/lib/postgresql/data
     environment:
-      - POSTGRES_PASSWORD=super_secret_password
+      - POSTGRES_PASSWORD=<PLEASE_ENTER_YOUR_PASSWORD>
       - POSTGRES_USER=db_user
       - POSTGRES_DB=stringer
 
   web:
-    image: mockdeep/stringer
+    image: mockdeep/stringer:latest
```
mockdeep, line 18: these images should actually move to `stringerrss/stringer` at some point. Not sure if now is the time, since we're currently [unable to publish the image just yet](https://app.circleci.com/pipelines/github/stringer-rss/stringer/1003/workflows/493f8194-6145-4321-b82e-a8b56056ebd0/jobs/1013).

docs/Docker.md
```diff
@@ -20,11 +20,12 @@ The following steps can be used to setup Stringer on Docker, using a Postgres da
...
     --restart always \
-    --volume /srv/stringer/data:/var/lib/postgresql/data \
+    --volume ./postgres:/var/lib/postgresql/data \
     --net stringer \
-    -e POSTGRES_PASSWORD=myPassword \
+    -e POSTGRES_PASSWORD=<PLEASE_ENTER_YOUR_PASSWORD> \
+    -e POSTGRES_USER=db_user \
     -e POSTGRES_DB=stringer \
-    postgres:9.5-alpine
+    postgres:12-alpine
```
mockdeep, line 28: How come you chose version `12` instead of the most recent `15`?

docs/Docker.md
```diff
@@ -20,11 +20,12 @@ The following steps can be used to setup Stringer on Docker, using a Postgres da
 docker run --detach \
     --name stringer-postgres \
     --restart always \
-    --volume /srv/stringer/data:/var/lib/postgresql/data \
+    --volume ./postgres:/var/lib/postgresql/data \
```
mockdeep, line 23: Can you explain this change to me? I'm not at all familiar with docker, so not sure why this would change.

mockdeep (commented):
@guidopetri do you have time to look this over?

docs/Docker.md
Ni-g-3l: I had an issues with postgres 9 which update all my folder permissions. Apparently it was fixed in 12 version, and it's the first version with the fix. May be you can use 15 version, but I haven't try it. 

I could do some test if you want !

mockdeep: @Ni-g-3l sure, that would be helpful! I'm running 15.3 with my instance. It's not a huge deal, though, we can always upgrade more later.

docs/Docker.md
Ni-g-3l: I thought it would be better to gives to people a command which isn't install things all around your disks. Like that a postgres folder will be created to handle the docker volume at the current working directory.

I can revert this part if you want !

Ni-g-3l: I thought it would be cleaner to create the volume on current working dir, where the user will run the command, it simplify the install, and if the user want remove stringer from his instance he just have to remove the folder.

docs/Docker.md
Ni-g-3l: I just upgraded my instance, let's pass this week and see what will happen.

Ni-g-3l: First issues incomming, you can't point to same folder postgres12 and postgres15 
image

guidopetri: I think this is more because of an upgrade of postgres using the same data directory, rather than a problem with postgres 15 itself. So updating from 9 to 12 is already going to give issues, I think we may as well do 15.

I think this is probably where a stringer export comes in - we could export the data, upgrade to a newer version of postgres, and then re-import.

imo, we should set this to 15. If a new user comes to stringer, they'll want the latest version of postgres; if a previous user is updating, they should know that they need to back up their system and that changing postgres versions incurs problems. We may want to announce this in some way in the next release though, as a PSA.

guidopetri: (also, fwiw, someone else I know from the FOSS community created [this repo](https://github.com/pgautoupgrade/docker-pgautoupgrade) that auto-upgrades postgres without issues like this, and then runs the container as postgres normally would. But it's an additional layer of complexity, and I imagine postgres will eventually coalesce things into their docker image)

docs/Docker.md
guidopetri: Existing users will need to move their directory from `/srv/stringer/data` to `./postgres`, yeah. I think this is a better default, though.

docker-compose.yml
```diff
@@ -1,27 +1,40 @@
...
+    hostname: stringer-postgres
     restart: always
+    networks:
+      - stringer
     volumes:
-      - ~/stringer:/var/lib/postgresql/data
+      - ./postgres:/var/lib/postgresql/data
     environment:
-      - POSTGRES_PASSWORD=super_secret_password
+      - POSTGRES_PASSWORD=<PLEASE_ENTER_YOUR_PASSWORD>
```
guidopetri, line 13: Personally I don't care too much for changes like these since the password will need to be changed anyway. If we want to make it so that the container won't launch without a password change then we should probably just not set this value

docs/Docker.md
guidopetri: I noticed in the docker-compose that the default dir there is `~/stringer` which imo is an even better default.

docker-compose.yml
```diff
@@ -1,27 +1,40 @@
...
+      - stringer-network
     environment:
-      - SECRET_KEY_BASE=<your configuration>
-      - ENCRYPTION_PRIMARY_KEY<your configuration>
-      - ENCRYPTION_DETERMINISTIC_KEY=<your configuration>
-      - ENCRYPTION_KEY_DERIVATION_SALT=<your configuration>
+      - SECRET_KEY_BASE=$(openssl rand -hex 64)
+      - ENCRYPTION_PRIMARY_KEY=$(openssl rand -hex 64)
+      - ENCRYPTION_DETERMINISTIC_KEY=$(openssl rand -hex 64)
+      - ENCRYPTION_KEY_DERIVATION_SALT=$(openssl rand -hex 64)
```
guidopetri, line 31: I'm not a huge fan of this - this will generate new keys every time. So if someone accidentally re-ups the container, I think it ends up generating new keys and so their data is now locked behind (lost) encryption keys.

docker-compose.yml
```diff
@@ -1,27 +1,40 @@
...
   web:
-    image: mockdeep/stringer
+    image: mockdeep/stringer:latest
     build: .
     depends_on:
       - postgres
     restart: always
     ports:
-      - 80:8080
+      - '127.0.0.1:8080:8080'
```
guidopetri, line 24: This will only work for users who have a reverse proxy on their machine; it's safer this way (and imo should be the way to go) but it's something to be aware of when merging this in. The docker compose file doesn't actually have a reverse proxy listed in it.

Ni-g-3l:
Hey ! I followed up your recommendation and I made changes. 

Feel free to comment If I missed something

docker-compose.yml
mockdeep: I'm fine with cosmetic changes, and this seems like maybe an improvement. If there's a way to force the user to input it, though, so that they can't miss it, that seems even better.

docs/Docker.md
mockdeep: Hmm, I'm not sure what the best approach here is. I'm imagining the upgrade process would be smoother if it was more gradual, but that seems like a pretty lengthy process. That repo you linked could be really handy to smooth that process.

> We may want to announce this in some way in the next release though, as a PSA.

I unfortunately haven't figured out a good release process yet. Maybe we need to set up a changelog and tagging somehow. It's a little weird, though, since we aren't releasing a gem or the like.

docker-compose.yml
```diff
@@ -1,27 +1,40 @@
...
       - ENCRYPTION_PRIMARY_KEY<your configuration>
       - ENCRYPTION_DETERMINISTIC_KEY=<your configuration>
       - ENCRYPTION_KEY_DERIVATION_SALT=<your configuration>
+      # - FETCH_FEEDS_CRON="*/5 * * * *" # optional
+      # - CLEANUP_CRON="0 0 * * *" # optional
       - PORT=8080
-      - DATABASE_URL=postgres://db_user:super_secret_password@postgres:5432/stringer
+      - DATABASE_URL=postgres://db_user:<PLEASE_ENTER_YOUR_PASSWORD>@stringer-postgres:5432/stringer
+
+networks:
```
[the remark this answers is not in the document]
mockdeep: That sounds reasonable. I think the network updates might be a good 4th PR, too.

docker-compose.yml
```diff
@@ -1,27 +1,40 @@
...
       - postgres
     restart: always
     ports:
-      - 80:8080
+      - '8080:8080'
+    networks:
+      - stringer-network
     environment:
       - SECRET_KEY_BASE=<your configuration>
       - ENCRYPTION_PRIMARY_KEY<your configuration>
```
gui-ying233, line 29: `- ENCRYPTION_PRIMARY_KEY=<your configuration>`: Missing a `=`, btw

Ni-g-3l:
I think this PR is no longer needed since Stringer already use Postgres 16. I think we can close it

Issue #1051, opened 2023-07-06
Docker fixes - new compose (podman support also)


hozza:
Potentially helping docker related issues #741 and #523 

## Repo Linking

There are several links under different usernames/repos. I'm not sure how mantainers want to manaage this but it would help with docker issues if there could be one official repo that's linked everywhere and also published on on Docker Hub under this name also.
 - "https://github.com/stringer-rss/stringer/" links to "https://github.com/swanson/stringer" under the repos about section.
 - The [docker docs](https://github.com/stringer-rss/stringer/blob/main/docs/docker.md) reference "stringer-rss/stringer" (presumably expanding to "docker.io/stringer-rss/stringer")
 - The [docker-compose file](https://github.com/stringer-rss/stringer/blob/main/docker-compose.yml) references "mockdeep/stringer" (presumably expanding to "docker.io/mockdeep/stringer")
 
 ## Compose Tweaks

Fixes and improvement suggestion plus support rootless containers, (podman, podman-compose), if you're happy with these suggestions I'll submit PR. These tweaks make the docker docs accurate also.
 
```yaml
services:
  stringer-postgres:
    image: docker.io/postgres:9.5-alpine # full registry paths for non-docker compatibility/verbosity 
    container_name: stringer-postgres # helps with podman-compose
    restart: unless-stopped
    volumes:
      - stringer:/var/lib/postgresql/data # use volume rather than mount point
    environment:
      - PUID=1000 # perms issues for the old ~/ vol mount
      - PGID=1000
      - POSTGRES_USER=db_user
      - POSTGRES_DB=stringer
      - POSTGRES_PASSWORD=super_secret_password # TODO change me

  stringer:
    image: docker.io/mockdeep/stringer # maybe stringer-rss/stringer? # full registry paths for non-docker compatibility/verbosity 
    container_name: stringer # helps with podman-compose
    depends_on:
      - stringer-postgres
    restart: unless-stopped
    ports:
      - 8080:8080 # not on privileged port for rootless
    environment:
      - PORT=8080
      - DATABASE_URL=postgres://db_user:super_secret_password@stringer-postgres:5432/stringer  # TODO change me
#      - SECRET_KEY_BASE=<your configuration>  # TODO change me
#      - ENCRYPTION_PRIMARY_KEY<your configuration>  # TODO change me
#      - ENCRYPTION_DETERMINISTIC_KEY=<your configuration>  # TODO change me
[...]

PR #1080, opened 2023-08-20
Build Docker image and push to Dockerhub

guidopetri:
As mentioned in [#1051](https://github.com/stringer-rss/stringer/issues/1051#issuecomment-1679387966), the CircleCI docker build step isn't working, and we want to move it to a Github Action instead. This PR has two commits:

- first, revert the CircleCI additions so that the CircleCI tests file is clean (via reverting the two commits from #1065)
- second, add a build+push to dockerhub github action.

This will require two secrets that I *believe* can be set in the repository settings somewhere: one for a `DOCKERHUB_USERNAME` and another for a `DOCKERHUB_TOKEN`. I also went ahead and added both an amd64 and an arm64 build here, and am hoping that it just works (tm).

I also once again set the build and push to only happen when `main` is pushed to / gets new commits, but I think we can change that to work only on this branch from my fork if we want to test before merging.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,56 @@
+name: Build Docker image and push to Dockerhub
+
+on:
+  push
+env:
+  IMAGE_NAME: stringerrss/stringer
+
+jobs:
+  buildx:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Extract Docker sha tag
+        id: get-tag-sha
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ env.IMAGE_NAME }}
+          tags: type=sha
+
+      - name: Extract Docker latest tag
+        id: get-tag-latest
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ env.IMAGE_NAME }}
+          tags: type=raw, value=latest
+
+      - name: Log in to Docker Hub
+        if: ${{ github.ref_name == 'main' }}
+        id: login
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and Push
+        id: build-and-push
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.ref_name == 'main' }}
+          tags: |
+            ${{ steps.get-tag-latest.outputs.tags }}
+            ${{ steps.get-tag-sha.outputs.tags }}
+          cache-from: type=gha
+          cache-to: type=gha
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,10 +3,9 @@ version: 2.1
 orbs:
   browser-tools: circleci/browser-tools@1.2.3
   node: circleci/node@5.0.0
-  docker: circleci/docker@2.2.0
 
 jobs:
-  run-tests:
+  build:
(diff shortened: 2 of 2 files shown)


guidopetri:
@mockdeep hopefully it works this time!

.github/workflows/build.yaml
mockdeep, line 1: In terms of naming, I think I mostly see yaml files with the `.yml` extension.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,57 @@
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and Push
+        id: build-and-push
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
```
mockdeep, line 52: I think we might be able to build on *every* Github push and only push to Docker only on `main` with something like:

```suggestion
          push: ${{ github.ref_name == 'main' }}
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,57 @@
+name: Build Docker image and push to Dockerhub
+
+on:
+  push:
+    branches:
+      - "main"
```
mockdeep, line 6: I'm wondering if we can have this build on *every* branch/commit and only push to Docker on `main`. That way we can be sure this builds, and subsequent changes don't break it, before merging.

```suggestion
on:
  push
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,57 @@
...
+        with:
+          images: ${{ env.IMAGE_NAME }}
+          tags: type=raw, value=latest
+
+      - name: Log in to Docker Hub
+        id: login
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
mockdeep, line 43: I've set these secrets, so hopefully that'll work.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,57 @@
...
+          tags: type=sha
+
+      - name: Extract Docker latest tag
+        id: get-tag-latest
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ env.IMAGE_NAME }}
+          tags: type=raw, value=latest
+
+      - name: Log in to Docker Hub
```
mockdeep, line 38: If we're building this on every branch, I think we'll need to skip this step unless we're on `main`:

```suggestion
      - name: Log in to Docker Hub
        if: ${{ github.ref_name == 'main' }}
```

mockdeep (commented):
Thanks so much for this! I think this will help a lot with environment issues.

.github/workflows/build.yaml
guidopetri: Renamed!

.github/workflows/build.yml
```diff
@@ -0,0 +1,56 @@
...
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.ref_name == 'main' }}
+          tags: |
+            ${{ steps.get-tag-latest.outputs.tags }}
+            ${{ steps.get-tag-sha.outputs.tags }}
+          cache-from: type=gha
```
guidopetri, line 55: I *believe* that this `cache-from` will cache to/from the Github Actions environment, so there's no need to pull an image locally, but I'm not 100% confident.

guidopetri:
Updated the PR with your suggestions and rebased them into the second commit. Should I keep the `run-tests` name on CircleCI instead of reverting to `build`?

mockdeep:
Let's give it a try! I'm wondering why we don't see the image building on this PR, but maybe we need to merge it before it will run.

> Should I keep the `run-tests` name on CircleCI instead of reverting to `build`

I think `build` is fine for now. We can always revisit later.

guidopetri:
It worked?!!

mockdeep:
Oh, I didn't even think to look. It did!

mockdeep:
https://hub.docker.com/repository/docker/stringerrss/stringer/general

PR #1082, opened 2023-08-22
Relocate Postgres Database and Upgrade to Postgres 12

Ni-g-3l:
Hey ! 

As mentionned [here](https://github.com/stringer-rss/stringer/pull/1071/files), I splitted my first PR into small one. As we can merge it easily !

In this one : 
- Sync Pg volume between documentation and docker-compose
- Upgrade to Postgres 12

Feel free to comment if I missed something

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     command: ["/app/docker/init_or_update_env.rb"]
 
   stringer-postgres:
-    image: postgres:9.5-alpine
+    image: postgres:12-alpine
     container_name: stringer-postgres
     restart: always
     depends_on:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     command: ["/app/docker/init_or_update_env.rb"]
 
   stringer-postgres:
-    image: postgres:12-alpine
+    image: postgres:16-alpine
     container_name: stringer-postgres
     restart: always
     depends_on:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ services:
     networks:
       - stringer-network
     volumes:
-      - ~/stringer:/var/lib/postgresql/data
+      - /srv/stringer/data:/var/lib/postgresql/data
     env_file: .env
 
   stringer:
--- a/docs/Docker.md
+++ b/docs/Docker.md
@@ -32,7 +32,7 @@ docker run --detach \
     --net stringer-network \
     -e POSTGRES_PASSWORD=myPassword \
     -e POSTGRES_DB=stringer \
-    postgres:9.5-alpine
+    postgres:12-alpine
 ```
 
 3. Run the Stringer Docker image:
--- a/docs/Docker.md
+++ b/docs/Docker.md
@@ -32,7 +32,7 @@ docker run --detach \
     --net stringer-network \
     -e POSTGRES_PASSWORD=myPassword \
     -e POSTGRES_DB=stringer \
-    postgres:12-alpine
+    postgres:16-alpine
 ```
 
 3. Run the Stringer Docker image:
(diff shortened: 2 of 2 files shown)


docs/docker.md
```diff
@@ -20,11 +20,11 @@ The following steps can be used to setup Stringer on Docker, using a Postgres da
 docker run --detach \
     --name stringer-postgres \
     --restart always \
-    --volume /srv/stringer/data:/var/lib/postgresql/data \
+    --volume ~/stringer:/var/lib/postgresql/data \
```
mockdeep, line 23: I honestly don't know anything about best practices with this sort of thing. Can you help me understand what the difference is here, and why this is an improvement?

docs/docker.md
```diff
@@ -20,11 +20,11 @@ The following steps can be used to setup Stringer on Docker, using a Postgres da
...
     --net stringer \
     -e POSTGRES_PASSWORD=myPassword \
     -e POSTGRES_DB=stringer \
-    postgres:9.5-alpine
+    postgres:12-alpine
```
mockdeep, line 27: We discussed briefly on your other PR, but do you have any sense of what steps a user will need to take in order to preserve the data in their Postgres instance while upgrading?

guidopetri:
@Ni-g-3l are you still planning on finishing this PR? if not, do you mind if I take over the postgres upgrade?

Ni-g-3l:
Hello sorry I've lost my dev env I won't be able to work on stringer anymore for now. I think you could take this PR If you want 😁no problem for me ✌️

docs/docker.md
guidopetri: I've reverted this change. I don't think the postgres data location on the host machine really matters; as long as it's somewhere specific to stringer that won't get accidentally overwritten/lost, it should be fine. I think a location in `~` is too easy to accidentally remove, so I think we're better served by keeping the `/srv` subdir. I've also seen usage of the `/opt` directory, but again, I think this is just personal choice.

docs/docker.md
guidopetri: The most straightforward way to upgrade would be:
- `docker -it -u postgres stringer-postgres pg_dumpall > pg_dump.bak`
- `mv /srv/stringer/data /srv/stringer/data_backup`
- `docker compose up -d` (this will rebuild the postgres container with the new version of postgres, and will start from scratch because there's no `/srv/stringer/data` directory anymore)
- `docker cp pgdump.bak stringer-postgres:/pgdump.bak`
- `docker exec -it stringer-postgres chown postgres:postgres /pgdump.bak`
- `docker exec -u postgres -it stringer-postgres psql -f /pgdump.bak postgres`

Essentially, dumping the database to a file + creating a backup, recreating the database docker container, then restoring the database from file. [According to the postgres docs](https://www.postgresql.org/docs/current/upgrading.html), this is "The traditional method for moving data, [...] though this can be slow." In our use case, I don't think this matters that much; stringer doesn't have enough data stored in the db to really make it a slow operation. The reasoning for following this upgrade path is because major versions of postgres can contain changes to the underlying data structure on disk, and so we cannot copy the data directory over directly, since the structure may have (and, going from 9.5 to 16, most probably has) changed.

An alternate upgrade path, also described in the postgres docs, is using `pg_upgrade`. I haven't used this personally, but this seems to require a lot more work with installing multiple binaries for postgres in a single host/container. I think that if an end user knows about this upgrade path, they most probably know how to use it, or can figure it out for themselves, so I don't think we need to concern ourselves with this.

What do you think? We can add the above commands to docs, or make this into its own release with upgrade instructions?

mockdeep: @guidopetri that sounds reasonable to me. Is there likely to be any issue with exporting and reimporting data across so many versions?

guidopetri: No; from the docs linked above as well: 

> Current releases of the dump programs can read data from any server version back to 9.2.

PR #1084, opened 2023-08-22
Create Stringer Network

Ni-g-3l:
Hey ! 

As mentionned [here](https://github.com/stringer-rss/stringer/pull/1071/files), I splitted my first PR into small one. As we can merge it easily !

In this one : 
- Create stringer network in docker-compose

Feel free to comment if I missed something


docker-compose.yml
```diff
@@ -18,10 +21,17 @@ services:
...
       - ENCRYPTION_PRIMARY_KEY<your configuration>
       - ENCRYPTION_DETERMINISTIC_KEY=<your configuration>
       - ENCRYPTION_KEY_DERIVATION_SALT=<your configuration>
       - PORT=8080
       - DATABASE_URL=postgres://db_user:super_secret_password@postgres:5432/stringer
+
+networks:
+  stringer-network:
+    external: false
+    name: stringer-network
```
mockdeep, line 37: Can you help me to understand what this change does and why it's an improvement?

mockdeep (commented):
Thanks so much for breaking these up @Ni-g-3l! I appreciate the contributions.

docker-compose.yml
Ni-g-3l: it's ensure when you create your stringer from docker-compose file to create dedicated network to stringer, containing the postgres container and the web app. This prevent stringer web app to try to connect to another postgres container (for nextcloud or joplin for example). 

Like this you can just throw your docker compose in portainer stack and just run it. Independantly how many postgres container you have in your local docker It will work because their are on the same network

mockdeep: Thanks for the explanation! This doesn't sound like it's going to cause any user facing impact. Is that right, or are they going to have to make some change on their end?

Ni-g-3l: No,it seems no. if new user follow documentation. It will create corresponding network and then if they drop the docker compose in some app like Portainer it will create the corresponding network too

Ni-g-3l:
Hey ! I renamed  docker doc file to keep constistency in documentation file name but I can't Fix the conflicts between the main branch and mine.

mockdeep:
@Ni-g-3l unfortunately, only you can push to your fork. I think if you merge main and then reset it should work. You'll need to write your commit message again:

```sh
git merge upstream/main
git reset upstream/main
git commit -m "create stringer network"
git push --force-with-lease
```

Ni-g-3l:
> @Ni-g-3l unfortunately, only you can push to your fork. I think if you merge main and then reset it should work. You'll need to write your commit message again:
> 
> ```shell
> git merge upstream/main
> git reset upstream/main
> git commit -m "create stringer network"
> git push --force-with-lease
> ```

It worked ! Thanks !

mockdeep:
Thank you!

PR #1089, opened 2023-08-26
precompile assets in the dockerfile before launching. closes #1086

### LINKED ISSUE #1086: Assets don't work without precompilation
Upon deploying the latest version of the repo via Docker, the CSS/JS assets don't load properly. Running `rails assets:precompile` and restarting the container fixed it, so we should do that as part of the Dockerfile.

guidopetri:
I added an asset precompilation step to the Dockerfile. However, this didn't work out of the box - it seems like [rails needs secret keys/environment variables to be loaded](https://github.com/rails/rails/issues/32947) to run any command, even just an asset precompilation. So I added some dummy arguments to the Dockerfile that are only available during build, not during runtime. Unfortunately, this seems to be the standard: e.g. [Mastodon](https://github.com/mastodon/mastodon/blob/main/Dockerfile#L100) does this.

This wasn't quite satisfactory, so I went with another option: to precompile assets in `docker/start.sh`, i.e. the entrypoint of the container. This is also a little inconvenient, because if someone changes the entrypoint to the container, then they might not precompile the assets and their image might be broken. But this seemed a lot cleaner than the following:

```Dockerfile
...
ADD . /app
# required for assets precompilation
ARG SECRET_KEY_BASE='dummy'
ARG ENCRYPTION_PRIMARY_KEY='dummy'
ARG ENCRYPTION_DETERMINISTIC_KEY='dummy'
ARG ENCRYPTION_KEY_DERIVATION_SALT='dummy'
RUN rails assets:precompile
```

I also tried fetching env vars with defaults in the `config/application.rb`, but the `config/initializers/dotenv.rb` prevented me from starting the container without the env vars actually set. (side note, I think we can deprecate `SECRET_KEY_BASE` now?)

### DIFF
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -7,6 +7,8 @@ if [ -z "$DATABASE_URL" ]; then
   exit 1
 fi
 
+rails assets:precompile
+
 : ${FETCH_FEEDS_CRON:='*/5 * * * *'}
 : ${CLEANUP_CRON:='0 0 * * *'}
 


guidopetri:
^The above all being said, I still get this sometimes:

image

with a lot of 404's on my reverse proxy, but no logs on the stringer container. So I'm not sure what's up yet.

guidopetri:
Oh, I think that's because of the other changes we made today that I didn't apply. I'm having some trouble getting the right combo again but I'll give it a better shot tomorrow, and eagerly await the other fixes to be committed, heh

docker/start.sh
```diff
@@ -7,6 +7,8 @@ if [ -z "$DATABASE_URL" ]; then
   exit 1
 fi
 
+rails assets:precompile
```
mockdeep, line 10: Unless I'm misunderstanding, it looks like this compiles assets when the app boots as opposed to when the image is built. That might be a bit of a performance annoyance if we were at Mastodon scale, but we've got a little ways to go, so I think this will do for now.

One other possibility would be to remove the `Dotenv` requirement, but that would make it less clear when someone is missing an essential key.

mockdeep (approved):

docker/start.sh
guidopetri: > it looks like this compiles assets when the app boots as opposed to when the image is built.

Yeah, that's correct.

I'm not sure what the consequences are of removing the `Dotenv` requirement, but to make the precompilation work in the Dockerfile I also had to use default values in `config/applications.rb`, which (if we did both) would mean that someone could theoretically run the application unsafely with the defaults for their keys. Personally I'd rather play it safe than sorry and make those env vars a hard requirement.

mockdeep: Yeah, or we could change the `.fetch` to instead be regular hash access, which won't raise an error:

```diff
-ENV.fetch('my_var')
+ENV['my_var']
```

But I agree, the hard requirement is probably better. Let's run with this for now.

PR #1091, opened 2023-08-26
Use env files instead of environment variables declared in docker-compose. Closes #1088

### LINKED ISSUE #1088: Convert docker compose to use env file + add env file creator
If we move the environment variables to an env file, we can set up a command like `make env` in the Makefile to create an env file for us without having to manually change values in the docker compose file (and without worrying about leaking secrets).

guidopetri:
It's best practice to leave secrets and environment variables outside of a docker compose file, because if there are secrets we can't add the file to a VCS.

This PR adds a new entrypoint to the Docker container to create required env variables for postgres and stringer if they don't already exist. It also changes the `docker-compose.yml` file to use these env files instead of environment variables passed in one by one. I also renamed the containers from `postgres` and `web` to `stringer-postgres` and `stringer` which are more identifiable in a setup that may have multiple services or applications running in docker.

Finally, I also updated the docker compose docs and upgraded the docker compose file version to 3.

The new instructions for easy setup via docker compose are:

- clone the repo
- `touch .env`
- `docker compose up -d`
- visit `localhost`


Makefile
```diff
@@ -0,0 +1,7 @@
+-include .postgres.env
+
+postgres_env:
+	echo "POSTGRES_USER=stringer\nPOSTGRES_PASSWORD=`pwgen -s 20`\nPOSTGRES_HOSTNAME=stringer-postgres\nPOSTGRES_DB=stringer\n" >> .postgres.env
```
mockdeep, line 4: I don't seem to have `pwgen` installed on my machine. Would it make sense to use `openssl rand` here, too?

Makefile
```diff
@@ -0,0 +1,7 @@
...
+
+stringer_env:
+	echo "SECRET_KEY_BASE=`openssl rand -hex 64`\nENCRYPTION_PRIMARY_KEY=`openssl rand -hex 64`\nENCRYPTION_DETERMINISTIC_KEY=`openssl rand -hex 64`\nENCRYPTION_KEY_DERIVATION_SALT=`openssl rand -hex 64`\n\nDATABASE_URL=postgres://`echo ${POSTGRES_USER}`:`echo ${POSTGRES_PASSWORD}`@`echo ${POSTGRES_HOSTNAME}`/`echo ${POSTGRES_DB}`\nFETCH_FEEDS_CRON='*/5 * * * *'\nCLEANUP_CRON='0 0 * * *'\n" >> .stringer.env
```
mockdeep, line 7: If we go the Ruby/rake way, we could also use erb to generate these files. E.g.:

```erb
# docker/.stringer.env.erb
SECRET_KEY_BASE=<%= `openssl rand -hex 64`.strip %>
ENCRYPTION_PRIMARY_KEY=<%= `openssl rand -hex 64`.strip %>
ENCRYPTION_DETERMINISTIC_KEY=<%= `openssl rand -hex 64`.strip %>
ENCRYPTION_KEY_DERIVATION_SALT=<%= `openssl rand -hex 64`.strip %>

DATABASE_URL=postgres://<%= "#{ENV['POSTGRES_USER']}:#{ENV['POSTGRES_PASSWORD']}@#{ENV['POSTGRES_HOSTNAME']}/#{ENV['POSTGRES_DB']}" %>
FETCH_FEEDS_CRON='*/5 * * * *'
CLEANUP_CRON='0 0 * * *'
```

I think the rake task would look something like:

```rb
namespace :env do
  task :generate_stringer do
    File.write("./.stringer.env", `erb ./docker/.stringer.env.erb`)
  end
end
```

That way it's a little more consistent with Rails conventions. Being a Ruby guy, I'm a bit biased, but I also find that to be a bit more readable.

docker-compose.yml
```diff
@@ -1,27 +1,19 @@
-version: '2'
+version: '3'
 services:
-  postgres:
+
+  stringer-postgres:
```
mockdeep, line 4: Do these names need to change to account for multiple apps running on the same machine?

Makefile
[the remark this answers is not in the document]
guidopetri: Sure thing! I wasn't aware of that, thanks for pointing it out.

Makefile
guidopetri: That was what I had initially, but the `-hex` flag (obviously) only returns hexadecimal characters, and `-base64` has the typical `==` ending that a lot of base64 strings have. It also seemed like I couldn't control the exact number of characters like with `pwgen`. I can look into a few more options.

Makefile
guidopetri: I agree, the `\n` delimited stuff isn't super readable. I could split it into multiple echo calls, but as you've mentioned, `rake` seems to be the standard so let's go with that :) though given my inexperience with Ruby there might be a few more comments from your part to make everything look right, heh.

docker-compose.yml
guidopetri: Technically they don't *need* to change, but yes - if you launch the current docker compose, it'll create two containers nicknamed `postgres` and `web`, but those names aren't very descriptive. In my machine I have 14 different containers running, and one "central" postgres container, so having them named properly / more descriptively is helpful. I think this was also part of someone else's suggestion (maybe @Ni-g-3l ?).

guidopetri:
I'll take a second pass through this PR and make the Rakefile modifications.

Makefile
mockdeep: So the issue with `-hex` is the limited character set? I'm not finding any clear message as to whether that is a problem. Looking at Heroku database urls, it looks like they're using hex characters on their end, so that's a vote of confidence. (They seem to prefix it with a `p` in some cases.)

guidopetri: Well, given it's a password, restricting the candidate character set makes it a lot easier to guess.

mockdeep: You made me go and learn about password strength. According to [Wikipedia](https://en.wikipedia.org/wiki/Password_strength):

> Due to currently understood limitations from fundamental physics, there is no expectation that any digital computer (or combination) will be capable of breaking 256-bit encryption via a brute-force attack.

Which may be the reason that 64 digit hex passwords are used, since 16^64 == 2^256, which achieves the highest recommended password entropy. That comes out to 115792089237316195423570985008687907853269984665640564039457584007913129639936 possible variations.

guidopetri: Hey, looking at Keepass for quality is actually a great idea. Here's 20 hex characters: 
image

I tried a few times and it ranged from 85-100 bits of entropy. According to [this chart](https://i.imgur.com/gfYw57t.png) that's still ~3 million years at the fastest attempts/s to crack it, on average, so I think I agree with you that it's probably enough for a default. Now to figure out how to make it yield 20 characters, though :D

guidopetri: Hey, looking at Keepass for quality is actually a great idea. Here's 20 hex characters:

image

I tried a few times and it ranged from 70-75 bits of entropy. According to [this chart](https://i.imgur.com/gfYw57t.png) that's 4.5 months at the fastest attempts/s to crack it, and half as long on average, so I think we'd have to make it longer still. 25 characters is 100 bits of entropy, so that should be enough (~200 million years to crack on average).

(NB: originally I had posted with an image that had non-hex characters and reached a different conclusion)

mockdeep: Heroku uses 64 characters for the db password, so I think we could probably run with that. `openssl rand -hex 32` generates a 64 character string.

docker-compose.yml
Ni-g-3l: Yes,  I agree, I have something like 20 containers on same instance, and many of them are database for service. So I think it's good practice to prefix your services with the app name. It make the instance maintenance process easier.

Rakefile
```diff
@@ -33,3 +33,17 @@ task :cleanup_old_stories, [:number_of_days] => :environment do |_t, args|
...
+desc "Generate Postgres env file"
+namespace :env do
+  task :generate_postgres do
+    File.write("./.postgres.env", `erb .docker/.postgres.env.erb`)
+  end
+end
+
+desc "Generate stringer env file"
+namespace :env do
+  task :generate_stringer: :generate_postgres do
```
guidopetri, line 46: I'm not sure this syntax is correct, but I wanted to make `env:generate_stringer` depend on `env:generate_postgres`. I tried to follow the syntax in the second line [here](https://ruby.github.io/rake/doc/rakefile_rdoc.html#label-Multiple+Definitions).

For the Docker user, how does one run these rake commands? Which package do we need to install to have `rake` available?

guidopetri: (if `env:generate_stringer` depends on `env:generate_postgres`, does it automatically run the dependency? i.e. could we just run `rake env:generate_stringer`? What happens if the `.postgres.env` or `.stringer.env` files already exist, does erb recognize that?)

Makefile
guidopetri: Thanks for this code snippet! This made it a lot easier to make the changes.

guidopetri:
Yay! It's passing. Ready for your re-review @mockdeep - I also had a few questions above.

Rakefile
mockdeep: > For the Docker user, how does one run these rake commands? Which package do we need to install to have `rake` available?

This depends on the Ruby gems being installed. Then we can run `bundle exec rails env:generate_stringer`.

> if `env:generate_stringer` depends on `env:generate_postgres`, does it automatically run the dependency?

Yep, it'll execute the dependency first.

> What happens if the `.postgres.env` or `.stringer.env` files already exist, does erb recognize that?

Based on my testing locally, it looks like it will overwrite the file. Should we add a check for `File.exist?` and break?

Rakefile
```diff
@@ -33,3 +33,15 @@ task :cleanup_old_stories, [:number_of_days] => :environment do |_t, args|
   args.with_defaults(number_of_days: 30)
   RemoveOldStories.call(args[:number_of_days].to_i)
 end
+
+namespace :env do
+  desc "Generate Postgres env file"
+  task generate_postgres: :environment do
```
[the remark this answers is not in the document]
guidopetri: Rubocop complained when I didn't have it, but maybe I confused it with other errors? I'll give it a try

Rakefile
```diff
@@ -33,3 +33,15 @@ task :cleanup_old_stories, [:number_of_days] => :environment do |_t, args|
...
+    File.write("./.postgres.env", `erb .docker/.postgres.env.erb`)
```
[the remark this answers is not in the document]
guidopetri: Nice catch. I meant to add a `./` prefix

Rakefile
guidopetri: Hmm, depending on Ruby being installed isn't super satisfactory - the whole point of docker is not to need anything spcial installed... let me think about this one.

Yeah, we should not overwrite the file. I think if we use `file` instead of `task` it has that behavior, but I can try with `File.exist` as well.

Rakefile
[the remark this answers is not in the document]
guidopetri: Yeah, still failing: [CircleCI](https://app.circleci.com/pipelines/github/stringer-rss/stringer/1062/workflows/ab42d1a0-441b-459f-968f-9c0a70feffd4/jobs/1076)

Rakefile
mockdeep: > the whole point of docker is not to need anything special installed.

In this case, Ruby and the required gems should be packed in with the image, so I think it should be okay. These commands will be run at runtime inside the image, right?


> I think if we use `file` instead of `task` it has that behavior

Wow, I didn't know about the `file` tasks. What if we do something like:

```rb
file ".postgres.env" do
  ...
end

file ".stringer.env" do
  ...
end

task generate: [".postgres.env", ".stringer.env"]
```

Rakefile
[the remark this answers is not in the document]
mockdeep: Hmm, I don't know if that's a very useful cop. You can disable it at the bottom section of `.rubocop.yml` if you want.

Rakefile
```diff
@@ -33,3 +33,15 @@ task :cleanup_old_stories, [:number_of_days] => :environment do |_t, args|
   args.with_defaults(number_of_days: 30)
   RemoveOldStories.call(args[:number_of_days].to_i)
 end
+
+namespace :env do
+  desc "Generate Postgres env file"
+  task :generate_postgres do
+    File.write("./.postgres.env", `erb ./docker/.postgres.env.erb`)
```
mockdeep, line 40: Thinking about these files, do we really need them to be separate? What if we had a single `.env.production.local` file that includes both of these?

Side note, we're using the dotenv gem. I'm not sure how that interplays with any builtin `.env` support Docker has, but they have a [list of recommendations](https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use).

Rakefile
guidopetri: > These commands will be run at runtime inside the image, right?

No, we'd need them to run outside of the image (otherwise there's high potential for data loss).

Rakefile
guidopetri: I'll take a look through those recs.

We should have the env files separate imo because someone might not want to run our provided postgres container but rather their own (with other dbs in it). It also is probably better from a security standpoint, I think?

Rakefile
mockdeep: Hmm, well docker doesn't really have much in the way of dependencies, does it? It doesn't seem like `make` is necessarily guaranteed either, is it? Does that leave us with just shell scripting?

Rakefile
mockdeep: Hmm, I guess I'm not sure how all the pieces will fit together here. I'm not able to think of any security benefit to keeping them separate. Users can edit the file afterwords if they want to point to their own PG installation, right? By default, dotenv [won't override existing env variables](https://github.com/bkeepers/dotenv#why-is-it-not-overriding-existing-env-variables), so if they have it setup somewhere else that should take precedence.

I'm mostly concerned about people having to be aware of multiple `.env` files that might impact their installation. It seems easier to have it all in one place.

Rakefile
mockdeep: Alternatively, can we run the command inside the docker image and sync it outside?

guidopetri: > It doesn't seem like make is necessarily guaranteed either, is it? Does that leave us with just shell scripting?

Well, it's near-guaranteed. It's not Docker itself but rather the system someone is running docker on that would need `make` or `rake` or whatever command we decide on.

> can we run the command inside the docker image and sync it outside?

I don't think so... we would have to run the docker container with a given command, copy the file outside the container, then re-launch the container. And the first time we run the docker container it would have the "wrong" environment variables (because they aren't defined). This could be done but is less-than-ideal because generally dependencies aren't defined within a container but rather the container definition should be standalone.

Actually, as I was writing the above, maybe we _could_ set it up within the container, and just have the first command in the docker setup be something like `docker run --entrypoint=/bin/sh stringer -- rake:generate_env >> .env`, with `rake:generate` just printing output to stdout. So the instructions would be

```sh
docker run ... >> .env
docker compose up -d stringer
```

I think that'd be a satisfactory conclusion, what do you think? We'd keep the `Rakefile` commands and wouldn't need any external dependencies, not even `make`, on the bare metal system.

Rakefile
guidopetri: I guess that's true, as long as we don't have the PG variables required in Dotenv.

Rakefile
mockdeep: Could we maybe use a volume for it? My docker understanding is very limited, but I'm imagining adding it to the `docker-compose.yml` file, `.env:.env`. Then maybe we could add the command to generate the file `start.sh`.

One additional consideration is what happens on the off chance somebody is trying to use the docker image on a windows machine. It seems less likely they'll have the build tools we need.

guidopetri: I think that might work, but I believe we'd have to ask people to create an empty .env file otherwise docker compose complains. And if the file exists on the host then it won't be overwritten by a docker container, so I don't think the container would be able to write to the host disk.

I still think the best conclusion would be to have a separate entrypoint or separate command that we can run with the container and then export to the .env on the host disk - we wouldn't need anything other than docker on the host.

mockdeep: Notes from our face-to-face discussion: https://docs.google.com/document/d/1UP4N0szZdYyjXtA2RucE2MxqEm6ln8dhtHQTaXC-o7g/edit#heading=h.dedb9t7qkmm9

docker/init_or_update_env.rb
```diff
@@ -0,0 +1,20 @@
...
+required_env_vars = Set.new(["SECRET_KEY_BASE",
+"ENCRYPTION_PRIMARY_KEY",
+"ENCRYPTION_DETERMINISTIC_KEY",
+"ENCRYPTION_KEY_DERIVATION_SALT"])
+
+# set operation to get only env vars we need that don't exist yet
+new_env_var_keys = required_env_vars - existing_env_vars
+
+for new_env_var_key in new_env_var_keys do
+    # TODO: generate the default
```
guidopetri, line 16: @mockdeep I'm struggling to conceptualize how to generate the env vars here. Am I on the wrong track? This feels a little primitive to do this set operation, and I seem to remember something about dotenv/erb that would help us here.

mockdeep: Hmm, so we want to add any missing variables, and keep all of the existing ones, right? I assume we want to ignore anything that is already in the environment, whether in `.env` or elsewhere. So maybe we should load in the Rails environment, then grab just the missing keys and append those to the `.env` file. Something like:

```rb
require_relative "../config/environment"

required_env = {
  "SECRET_KEY_BASE" => `openssl rand -hex 64`.strip,
  "ENCRYPTION_PRIMARY_KEY" => `openssl rand -hex 64`.strip,
  "ENCRYPTION_DETERMINISTIC_KEY" => `openssl rand -hex 64`.strip,
  "ENCRYPTION_KEY_DERIVATION_SALT" => `openssl rand -hex 64`.strip,
}

required_env.each do |key, value|
  next if ENV[key].present?

  File.open(".env", "a") { |file| file << "#{key}=#{value}\n"
end
```

guidopetri: I don't know why this wasn't obvious to me, it's like my brain turned to mush as soon as Ruby got involved. Thanks lol. I didn't know you could use `openssl rand` like that within Ruby, that's neat.

guidopetri: Alright, now we're cooking! I was able to generate the `.env` file as discussed - we `touch .env`, then run `docker compose up stringer-setup`, and it writes any required-but-not-already-existing env vars to the env file in the host machine. Let me know if this Ruby script looks good? A couple of comments:
- I had to remove the `../config/environment` requirement because if the application was initialized without the required keys, it crashed
- I used `has_key?` instead of `present?` but I'm not sure what the difference is (if any). Is one more idiomatic than the other?
- I think the `DATABASE_URL` has to be added separately from the other vars, because it depends on the vars being generated. Unless we're able to reference hash keys/values that were already introduced, but before the hash was "finished"?
- For the `DATABASE_URL`, we need to get the actual values of the environment if they exist, so I used some ternaries to declare the db, username, password etc. Is this idiomatic? I saw something like `ENV[key] || default` as well that we could use, but that seems to have edge cases around nil/false. (maybe there's a different way of fetching a key from a hash with a default if it's not present, too?)

guidopetri:
@mockdeep should be ready for your re-review now, sorry for the delay.

guidopetri:
@mockdeep should be good for another review now.

mockdeep (approved):
@guidopetri looks great! Thanks for your persistence on this. Feel free to squash it and merge it when you're ready! Please include the PR description in the commit message.

guidopetri:
🎉 thanks for your patience and for all the reviews!

PR #1092, opened 2023-08-26
Use `TARGETARCH` to get supercronic binary per target platform

guidopetri:
I was looking into how to pass the target arch into a build arg in the Dockerfile, when I found out that it's already available in `TARGETARCH`. So instead of using `ARCH` I renamed everything to `TARGETARCH` and got a successful amd64 build and arm64 build, with the correct versions of the supercronic binary attached. [Here's](https://github.com/guidopetri/stringer/actions/runs/5985463167/job/16237585875) the GH Actions logs; if you search for `supercronic` you can see L738/739 shows the amd64 build + output, and L1197/1198 shows the arm64 build + output.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,13 +25,13 @@ RUN DEBIAN_FRONTEND=noninteractive dpkg-reconfigure locales \
 
 ENV LC_ALL C.UTF-8
 
-ARG ARCH=amd64
-ENV SUPERCRONIC_URL=https://github.com/aptible/supercronic/releases/download/v0.1.3/supercronic-linux-$ARCH \
-    SUPERCRONIC=supercronic-linux-$ARCH \
+ARG TARGETARCH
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -25,13 +25,13 @@ RUN DEBIAN_FRONTEND=noninteractive dpkg-reconfigure locales \
 
 ENV LC_ALL C.UTF-8
 
-ARG ARCH=amd64
-ENV SUPERCRONIC_URL=https://github.com/aptible/supercronic/releases/download/v0.1.3/supercronic-linux-$ARCH \
-    SUPERCRONIC=supercronic-linux-$ARCH \
+ARG TARGETARCH
```
mockdeep, line 28: How does `TARGETARCH` get populated? Is id thanks to [the `platform` line](https://github.com/stringer-rss/stringer/blob/5dd869fe2c850405e1903c42f9def98f37a0bb28/.github/workflows/build.yml#L50) in our Github action?

mockdeep (approved):

PR #1102, opened 2023-09-16
set RAILS_SERVE_STATIC_FILES in the docker container

guidopetri:
Without setting this variable, Rails won't serve static files from within the Docker container while in production mode. We could circumvent this by having a shared volume to the assets folder and serving the assets through a reverse proxy, but it seems more trouble than it's worth at this point.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,4 +47,6 @@ RUN useradd -m stringer
 RUN chown -R stringer:stringer /app
 USER stringer
 
+ENV RAILS_SERVE_STATIC_FILES=true
+
 CMD /app/start.sh


guidopetri:
@mockdeep merging this in per our chat yesterday

Dockerfile
```diff
@@ -47,4 +47,6 @@ RUN useradd -m stringer
 RUN chown -R stringer:stringer /app
 USER stringer
 
+ENV RAILS_SERVE_STATIC_FILES=true
```
mockdeep, line 50: @guidopetri when do we decide to put an env variable in the `Dockerfile` vs the `.docker-compose.yml`?

guidopetri: Generally any "default" things that don't need customization, or that customization significantly alters the inner workings of the app, should go in the Dockerfile imo. The env var can still be overwritten in the docker compose file though. So e.g. we might set the default port to 8080 in the Dockerfile, but then also overwrite it to 80 in the compose

PR #1115, opened 2023-11-02
Removed memoization from feed/create and feed/export_to_opml specs

RobBoothAppDev21:
Updated spec files below to address "MultipleMemoizedHelpers" cop

- spec/feeds/create_spec.rb
- spec/feeds/export_to_opml_spec.rb


mockdeep:
Thank you!

PR #1124, opened 2023-11-19
Update Fetch One spec

RobBoothAppDev21:
Introduce new xml generator and use that rather than stubbing private
method


mockdeep (commented):
Nicely done! If you update the `HTTParty` stubbing this should be good to go.

PR #1128, opened 2023-11-21
Fix install instructions

kimadactyl:
These are missing a lil step, and updated to the recent rails idiom

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Then run the following commands.
 
 ```sh
 bundle install
-rake db:migrate
+rails db:setup db:migrate
 foreman start
 ```
 


README.md
```diff
@@ -100,7 +100,7 @@ Then run the following commands.
 
 ```sh
 bundle install
-rake db:migrate
+rails db:setup db:migrate
```
mockdeep, line 103: Actually, I think these are mutually exclusive. If you're setting up a new environment `db:setup` makes sense. If you're working with an existing environment, you would want to `db:migrate` instead.

```suggestion
rails db:setup
```

mockdeep (commented):
Thanks for contributing @kimadactyl!

mockdeep:
@kimadactyl thanks for contributing! I've merged [another PR](https://github.com/stringer-rss/stringer/pull/1152) that changes it to `db:setup`.

PR #1129, opened 2023-11-23
Updated find new stories spec

RobBoothAppDev21:
Removed "let" and "before" from find_new_stories_spec


spec/commands/feed/find_new_stories_spec.rb
```diff
@@ -1,62 +1,60 @@
...
-
-      time = Time.zone.local(2013, 1, 2)
-      result = described_class.call(feed, 1, time)
-      expect(result).to eq([story2])
+      feed = create(:feed)
+      create(:story, entry_id: "www.story1.com", feed_id: feed.id)
+      entry = create_entry(published: Time.zone.now, url: "www.story2.com")
+      raw_feed = create_raw_feed(feed, entries: [entry])
+
+      expect(described_class.call(raw_feed, feed.id)).to eql(raw_feed.entries)
```
mockdeep, line 39: It's not a huge deal since they are equivalent in this case, but I generally us `eq` instead of `eql`. The [differences are subtle](https://stackoverflow.com/a/54088700/372479) and in many cases irrelevant, but for the sake of consistency I think it's probably worth sticking with `eq`.

```suggestion
      expect(described_class.call(raw_feed, feed.id)).to eq(raw_feed.entries)
```

spec/commands/feed/find_new_stories_spec.rb
```diff
@@ -1,62 +1,60 @@
...
-    )
-
-    result = described_class.call(feed, 1)
-    expect(result).not_to include(stories_older_than_3_days)
+    feed = create(:feed)
+    new_entry = create_entry(published: 2.days.ago)
+    old_entry = create_entry(published: 4.days.ago)
+    raw_feed = create_raw_feed(feed, entries: [old_entry, new_entry])
+
+    expect(described_class.call(raw_feed, feed.id)).not_to include(old_entry)
```
mockdeep, line 58: I think the existing test had a falsely passing test, since it was checking for the array `stories_older_than_3_days` rather than individual items. This highlights a reason I generally avoid things like checking negative inclusion, since the number of items *not* in the list is infinite. It's easy to get a test that is accidentally passing without really testing anything. So, I will instead test what *is* in the list, and often the *whole* list for the sake of precision:

```suggestion
    expect(described_class.call(raw_feed, feed.id)).to eq([new_entry])
```

mockdeep (commented):
A couple of minor comments. Over all, looks great, though. Thanks!

PR #1145, opened 2024-01-19
refactor: Removed Memoization From Commands Specs

RobBoothAppDev21:
Completed removal of remaining memoization across specs in commands folder.

Modified specs:
	-  spec/commands/feed/import_from_opml_spec.rb
	-  spec/commands/fever_api/read_favicons_spec.rb
	-  spec/commands/fever_api/write_mark_feed_spec.rb
	-  spec/commands/story/mark_as_read_spec.rb

### DIFF
--- a/spec/commands/feed/import_from_opml_spec.rb
+++ b/spec/commands/feed/import_from_opml_spec.rb
@@ -1,104 +1,107 @@
 # frozen_string_literal: true
 
 RSpec.describe Feed::ImportFromOpml do
-  let(:subscriptions) do
+  tmw_football = {
+    name: "TMW Football Transfer News",
+    url: "http://www.transfermarketweb.com/rss"
+  }
+
+  giant_robots = {
+    name: "GIANT ROBOTS SMASHING INTO OTHER GIANT ROBOTS - Home",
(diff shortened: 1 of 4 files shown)


spec/commands/feed/import_from_opml_spec.rb
```diff
@@ -1,104 +1,111 @@
...
+    url: "http://feeds.feedburner.com/GiantRobotsSmashingIntoOtherGiantRobots"
+  }
+
+  autoblog = { name: "Autoblog", url: "http://feeds.autoblog.com/weblogsinc/autoblog/" }
+  city_guide = { name: "City Guide News", url: "http://www.probki.net/news/RSS_news_feed.asp" }
+
+  def create_subscriptions(
+    file: "../../support/files/subscriptions.xml",
+    path: __dir__
+  )
```
mockdeep, line 20: In terms of naming, this seems more like a `subscriptions` or `read_subscriptions`. Also, I don't see `file` or `path` being passed in anywhere, so maybe they can be hard coded in the method?

```suggestion
  def read_subscriptions
```

mockdeep (commented):
Some minor comments. Thanks for contributing!

PR #1147, opened 2024-01-22
refactor: removed memoization from repository spec

RobBoothAppDev21:
Removed "let" and "before" from FeedRepository spec and StoryRepository spec

Files modified:
	spec/repositories/feed_repository_spec.rb
	spec/repositories/story_repository_spec.rb


mockdeep:
Great work, thanks!

PR #1166, opened 2024-03-04
FIX: Add default precision to db

RobBoothAppDev21:
Generate migration file to set the default precision to 6 for datetime columns in the database

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2023_08_01_025234) do
+ActiveRecord::Schema[7.1].define(version: 2024_02_26_201050) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -152,13 +152,13 @@
...
 
-  create_table "users", id: :serial, force: :cascade do |t|
-    t.string "password_digest", limit: 255
-    t.datetime "created_at", precision: nil, null: false
-    t.datetime "updated_at", precision: nil, null: false
-    t.string "api_key", limit: 255, null: false
-    t.string "username", null: false
+  create_table "users", force: :cascade do |t|
+    t.string "password_digest"
+    t.datetime "created_at"
```
mockdeep, line 157: Looks like we lost the `null: false` from some of these.

db/schema.rb
```diff
@@ -152,13 +152,13 @@
...
   end
 
-  create_table "users", id: :serial, force: :cascade do |t|
-    t.string "password_digest", limit: 255
-    t.datetime "created_at", precision: nil, null: false
-    t.datetime "updated_at", precision: nil, null: false
-    t.string "api_key", limit: 255, null: false
-    t.string "username", null: false
+  create_table "users", force: :cascade do |t|
+    t.string "password_digest"
```
mockdeep, line 156: Looks like this changed as well. Probably need a separate migration to remove the limit here.

PR #1172, opened 2024-03-16
Allow users to configure the news feed order

jhass:
I prefer to catch up with the older stories first and it always felt awkward going through the feed in reverse. So I always had a local patch to flip the order, decided to to push a proper solution upstream since I changed my deployment just now. Thanks for picking up the project btw!


spec/system/profile_spec.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 RSpec.describe "profile page" do
-  it "allows the user to edit their username" do
+  before do
```
mockdeep, line 4: We've been moving in the direction of eliminating `before` calls. I think it's fine to copy these over to these lines to the new spec.

spec/system/profile_spec.rb
```diff
@@ -35,4 +34,11 @@ def fill_in_password_fields(existing_password, new_password)
       fill_in("Password confirmation", with: new_password)
     end
   end
+
+  it "allows the user to edit their feed order" do
+    within_fieldset("Stories Feed Settings") { select("Oldest first") }
```
mockdeep, line 39:
```suggestion
    select('Oldest first', from: 'Stories Feed Order')
```

config/locales/en.yml
```diff
@@ -187,3 +187,10 @@ en:
     title: 'Stringer is '
     your_feeds: your feeds
     your_stories: your stories
+  activerecord:
+    attributes:
+      user:
+        stories_order: "Stories Feed Order"
```
mockdeep, line 193:
```suggestion
        stories_order: "Stories feed order"
```

app/views/profiles/edit.html.erb
```diff
@@ -1,12 +1,20 @@
 <h1><%= t('.title') %></h1>
 
-<header class="warning">
-  <p><%= t('.warning_html') %></p>
-</header>
+<%= form_with(model: user, url: profile_path) do |form| %>
```
mockdeep, line 3: I'm fine merging this where it is, but I think it would be nice if this was located on the stories page for easier access. We can always address that later, though.

app/controllers/stories_controller.rb
```diff
@@ -2,7 +2,8 @@
 
 class StoriesController < ApplicationController
   def index
-    @unread_stories = authorization.scope(StoryRepository.unread)
+    order = current_user.stories_order
+    @unread_stories = authorization.scope(StoryRepository.unread(order:))
```
mockdeep, line 6: I'm not super familiar with fever api, but should the queries in there be updated to match as well? Or do we expect the client to handle it instead?

mockdeep (commented):
Thank you for contributing this! I've been meaning to do this [for a while](https://github.com/stringer-rss/stringer/issues/159). I've also got it patched on my production instance.

I made some minor formatting suggestions, and added some architectural thoughts for your consideration. I'm fine not making major changes for now, though.

spec/system/profile_spec.rb
jhass: Your Rubocop config didn't allow me too, it was too many lines for what I needed. By the way I spent pretty much half the time fighting with Rubocop, I would recommend just using standardrb. I can imagine being too strict here can actually drive away contributors ;)

app/controllers/stories_controller.rb
jhass: I feel like that should be the clients responsibility. Unfortunately the original spec seems gone, but I doubt it specified any contract for this at all given the overall variance in the client implementations.

Given the fever API allows partial syncs with `since_id`, I can't imagine server side sorting to work reliably anyhow, imagine a sync history where the order is flipped between two partial syncs.

jhass: Yep, looking at [archive.org](https://web.archive.org/web/20221221112459/https://feedafever.com/api), there aren't even any response formats described, let alone semantics like item order.

spec/system/profile_spec.rb
mockdeep: Hmm, it works for me when I move it down. It's limited to 5 lines, and you've only got 3.

> I can imagine being too strict here can actually drive away contributors

Not having strict linters would drive _me_ away, and I'm [one of the most active contributor this project has ever had](https://github.com/stringer-rss/stringer/graphs/contributors). I'm happy to help others figure out how to gracefully appease them, though.

I'm not a fan of libraries like standardrb. They're not strict enough for my taste and not flexible enough.

mockdeep:
Going ahead and merging this. We can deal with the `before` another time.

PR #1176, opened 2024-03-23
fix: Factory story unread default and memoization

RobBoothAppDev21:
Update stories factories:
- Removed :unread trait from stories factory
- Removed :unread from create(:story) throughout specs

Removed memoization:
- spec/requests/stories_repository_spec
- spec/requests/tutorials_controller_spec

Updated rubocop_todo


mockdeep (approved):
Looks great!

PR #1212, opened 2024-07-17
Refactor migrations to not use ActiveRecord for compatibility

osbre:
Fixes https://github.com/stringer-rss/stringer/issues/1183.

### DIFF
--- a/db/migrate/20230223045525_add_null_false_to_associations.rb
+++ b/db/migrate/20230223045525_add_null_false_to_associations.rb
@@ -2,13 +2,21 @@
 
 class AddNullFalseToAssociations < ActiveRecord::Migration[7.0]
   def change
-    if User.any?
-      Feed.where(user: nil).update_all(user_id: User.first.id)
-      Group.where(user: nil).update_all(user_id: User.first.id)
-    end
+    update_null_foreign_keys
 
     change_column_null :feeds, :user_id, false
     change_column_null :groups, :user_id, false
     change_column_null :stories, :feed_id, false
+  end
+
+  private
+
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -13,6 +13,7 @@ class User < ApplicationRecord
 
   before_save :update_api_key
 
+  attribute :stories_order, :string
```
mockdeep, line 16: This should be a database attribute, so I don't think this will help. This creates an in-memory attribute, but will not persist to the database, so the stories order will not change.

For some reason, it appears that the migrations haven't run on the build images when somebody is trying to upgrade their Docker image. When I try setting up an image from scratch, I don't get the error.

osbre: Did you try running migrations from scratch?

mockdeep:
@osbre thanks for contributing! I think this will silence the issue, but it will also create a new one, since it will break the `stories_order` feature.

app/models/user.rb
mockdeep: All I've done so far is to follow [the docker instructions](https://github.com/stringer-rss/stringer/blob/5d0b9f65a6011bc9ebc4bd5535ed5924657e404b/docs/Docker.md) from our docs with a fresh environment. Unfortunately, I haven't had time to do any deeper investigation.

osbre:
Thanks @mockdeep! Indeed, you're right. Then I suggest removing ActiveRecord usage from migrations for compatibility. It would also future-proof any other breaking changes.

(I pushed the new changes)

mockdeep (approved):

PR #1229, opened 2024-09-15
fix #1213 adding a class

### LINKED ISSUE #1213: "Feed Url" label overlaps input on the add feed screen
see screenshot.

CleanShot 2024-07-19 at 22 41 49@2x

it looks ok before you try putting any content in it

CleanShot 2024-07-19 at 22 43 46@2x

I sanity checked that this wasn't just a large font size issue. It's not. 

The short width of the box is controlled by this css: 

```
.setup .control-group input {
  width: 210px;
}
```

which, in addition to not being wide enough, is problematic in that it is defined as i fixed pixel size and thus doesn't expand with the size of a user's font setting. Should be switched to `rem` instead of `px` 

the problem with the "Feed URL" text is caused by this CSS

```
.setup .field-label {
  display: block;
  color: #BFC9CA;
  position: absolute;
  left: 13px;
  top: 10px;
  -webkit-transition: 0.25s;
  -moz-transition: 0.25s;
  -o-transition: 0.25s;
  transition: 0.25s;
}
```
This is using an absolute pixel based positioning (again a problem as font sizes - and thus line height- grow).  It's interacting negatively with the input which has _NO_ absolute positioning. The small height forces it to be over (or under) the input. if you remove the `top: 10px` it drops down to below the input which makes sense given the HTML that's creating these 2 elements and it looks like this.

CleanShot 2024-07-19 at 22 52 33@2x

Given the `top: 10px` I'm unsure if it's supposed to be above or below the input. 

(tested using Firefox)

giovapanasiti:
it was just the missing class :)

### DIFF
--- a/app/views/feeds/new.html.erb
+++ b/app/views/feeds/new.html.erb
@@ -9,7 +9,7 @@
   <hr />
   <%= form_with(url: "/feeds", id: "add-feed-setup") do %>
     <div class="control-group">
-      <input name="feed_url" id="feed-url" type="text" value="<%= @feed_url %>" autofocus/>
+      <input name="feed_url" class="form-control" id="feed-url" type="text" value="<%= @feed_url %>" autofocus/>
       <i class="icon-rss field-icon"></i>
       <label id="feed_url-label" class="field-label" for="feed_url"><%= t('feeds.add.fields.feed_url') %></label>
     </div>


mockdeep:
Excellent, thanks for the fix @giovapanasiti!

PR #1266, opened 2025-01-08
Allow empty name during import and use FeedDiscovery to try to figure the name out if possible

renatolond:
While testing stringer locally, I tried to import my feeds from The Old Reader. The OPML generated by The Old Reader doesn't contain the feed names, so the import failed.
I made a change to allow nils and the import went through, but then the feeds had no name and it was hard to manage the feeds.

So I replicated the logic from `Feed::Create` to try to get a proper title for the Feed.

This will make the import slower in the case there's nameless entries, but the result seems to be better. I'm falling back to the url if the call to the feed failed to avoid retrying on dead feeds.

### DIFF
--- a/app/commands/feed/import_from_opml.rb
+++ b/app/commands/feed/import_from_opml.rb
@@ -30,15 +30,20 @@ module Feed::ImportFromOpml
     def create_feed(parsed_feed, group, user)
       feed = user.feeds.where(**parsed_feed.slice(:name, :url))
                  .first_or_initialize
-      if parsed_feed[:name].nil? && feed.name.nil?
-        result = FeedDiscovery.call(parsed_feed[:url])
-        title = result.title if result
-        parsed_feed[:name] = ContentSanitizer.call(title.presence || parsed_feed[:url])
-        feed.name = parsed_feed[:name] if parsed_feed[:name].present?
-      end
+      find_feed_name(feed, parsed_feed)
       feed.last_fetched = 1.day.ago if feed.new_record?
       feed.group_id = group.id if group
       feed.save
+    end
+
+    def find_feed_name(feed, parsed_feed)
+      return unless parsed_feed[:name].nil? && feed.name.nil?
+
+      result = FeedDiscovery.call(parsed_feed[:url])
+      title = result.title if result
+      parsed_feed[:name] =
+        ContentSanitizer.call(title.presence || parsed_feed[:url])
+      feed.name = parsed_feed[:name]
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/commands/feed/import_from_opml.rb
```diff
@@ -30,9 +30,20 @@ def find_or_create_group(group_name, user)
...
       feed = user.feeds.where(**parsed_feed.slice(:name, :url))
                  .first_or_initialize
+      find_feed_name(feed, parsed_feed)
       feed.last_fetched = 1.day.ago if feed.new_record?
       feed.group_id = group.id if group
       feed.save
     end
+
+    def find_feed_name(feed, parsed_feed)
+      return unless parsed_feed[:name].nil? && feed.name.nil?
```
mockdeep, line 40: It looks like these will be the same being that we're querying for `feed` based on the name from `parsed_feed`. Could we instead do:

```suggestion
      return if feed.name?
```

app/commands/feed/import_from_opml.rb
```diff
@@ -30,9 +30,20 @@ def find_or_create_group(group_name, user)
...
+
+      result = FeedDiscovery.call(parsed_feed[:url])
+      title = result.title if result
+      parsed_feed[:name] =
+        ContentSanitizer.call(title.presence || parsed_feed[:url])
+      feed.name = parsed_feed[:name]
```
mockdeep, line 46: It doesn't look to me like we need to assign the value back to `parsed_feed`. Can we do:

```suggestion
      feed.name = ContentSanitizer.call(title.presence || parsed_feed[:url])
```

mockdeep (commented):
A couple of minor comments. Thanks for contributing!

app/commands/feed/import_from_opml.rb
renatolond: It looks like it, but the `parsed_feed.slice(:name, :url)` is a bit misleading. If there's only one of the fields, it will only return that one:
```
3.4.1 :001 > parsed_feed = { name: "foo", url: "http://foo.com" }
 => {name: "foo", url: "http://foo.com"} 
3.4.1 :002 > parsed_feed.slice(:name, :url)
 => {name: "foo", url: "http://foo.com"} 
3.4.1 :003 > parsed_feed = { name: "foo" }
 => {name: "foo"} 
3.4.1 :004 > parsed_feed.slice(:name, :url)
 => {name: "foo"} 
3.4.1 :005 > parsed_feed = { url: "http://foo.com" }
 => {url: "http://foo.com"} 
3.4.1 :006 > parsed_feed.slice(:name, :url)
 => {url: "http://foo.com"} 
```
In other words, slice will not return nil fields, fields not found are just ignored.

I'll tweak the spec to make that clearer, I thought about it but forgot to do it

app/commands/feed/import_from_opml.rb
renatolond: Good catch, changed!

mockdeep:
Looks great! Thanks for the  contribution!

PR #1327, opened 2025-08-15
Do not render null if title is nil

bitboxer:
Newsflash on Gnome and some other apps really don't like if a title is null and crash. This prevents this.


mockdeep:
Thank you!

PR #1531, opened 2026-06-05
Heroku postgresql hobby-dev is no longer available

nickhammond:
Reference https://devcenter.heroku.com/articles/heroku-postgres-plans


mockdeep:
@nickhammond thanks for contributing!

